PR #15, opened 2020-10-22
stable object manager

MarioEnr:
Full object manager ready to use.
This is a manager that allows you to register any object to avoid being deleted and re instantiated by the network bubble and view distance. It also allows you to remove map objects taking into account all it's quirks (Collisions, making the renderer ignore them, etc).

For testing use 
`CF__ObjectManager.RemoveObjectsInRadius2D("1624.44 451.742 14180", 1000);`
In both client and server. this will remove a big chunk of Tisy Military.


Arkensor:
@Jacob-Mango I am reviewing this for a potential merge.

Arkensor:
Merged

@Jacob-Mango do not release until update 1.1 with the new structure.

PR #26, opened 2020-12-03
feature/ghosticons

Jacob-Mango:


InclementDab (approved):
Reviewed, tested and approved.

JM/CF/Defines/CFDefines.c
```diff
@@ -2,6 +2,8 @@
 #define CF_MODULE_CONFIG
 //#define CF_MODULE_PERMISSIONS
 
+#define CF_GHOSTICONS_V2
```
Arkensor, line 5: This needs to be changed to 
#define CF_GHOSTICONS

Arkensor (approved):
Approved with one minor comment that needs to be addressed post launch

PR #40, opened 2021-02-25
Feature/debugger

Jacob-Mango:


JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger_Block.c
```diff
@@ -0,0 +1,144 @@
...
+	void SwapBuffer()
+	{
+		GetTemplateController().EntryKeys = "";
+		GetTemplateController().EntryValues = "";
+
+		for (int i = 0; i < m_Entries.Count(); i++)
+		{
+			if (m_Entries[i].Key == "") m_Entries[i].Key = " ";
+			if (m_Entries[i].Value == "") m_Entries[i].Value = " ";
+			GetTemplateController().EntryKeys = GetTemplateController().EntryKeys + m_Entries[i].Key + "\n";
```
Arkensor: GetTemplateController().EntryKeys is fetched twice, but wasn't it overridden in line 93? So shouldn't it just be "" or delete completly?

JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger_Block.c
```diff
@@ -0,0 +1,144 @@
...
+	{
+		m_Buffered = buffered;
+		m_SwapClears = swapClears;
+	}
+
+	override string GetLayoutFile()
+	{
+		return "JM/CF/Debugger/layouts/Debugger_Block.layout";
+	}
+};
```
Arkensor: Please add one empty line at the end of each text file for git. e.g.

```
class Name
{
};

```

JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger.c
```diff
@@ -0,0 +1,199 @@
...
+
+		m_IsEnabled = FileExist("$profile:CF_Debugger.txt");
+		if (GetGame().IsServer() && GetGame().IsMultiplayer()) m_IsEnabled = false;
+	}
+
+	/*
+		Retrieves the Debugger block if it exists based on the name and if specified, the target. 
+		Creates a new block if no existing block was found.
+	*/
+	Class Get(string name, Object target = null)
```
Arkensor: Why is class returned and not a block type?

JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger.c
```diff
@@ -0,0 +1,199 @@
...
+
+		auto block = new CF_Debugger_Block(name, target);
+		m_Blocks.Insert(block);
+		return block;
+	}
+
+	/*
+		Get all blocks by name, if no name specified, get all blocks
+	*/
+	array<Class> GetAll(string name = "")
```
Arkensor: Same as above, why an array of classes and not an array of blocks

JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger.c
```diff
@@ -0,0 +1,199 @@
...
+class CF_Debugger
+{
+	private autoptr array<ref CF_Debugger_Block> m_Blocks;
+	private bool m_IsEnabled;
+
+	void CF_Debugger()
+	{
+		m_Blocks = new array<ref CF_Debugger_Block>();
+
+		m_IsEnabled = FileExist("$profile:CF_Debugger.txt");
```
Arkensor: Why does it need this file? There should be a way to enable and disable this purely via script and or startup parameter?

e.g. DayZ.exe -CF_DebugUI

JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger.c
```diff
@@ -0,0 +1,199 @@
+/* Example:
+
+void EOnSimulate(...)
+{
+	auto dbg_Helicopters = CF.Debugger.Get("Helicopters", this);
```
Arkensor: I am unsure what this returns. Does it return the block called helicopters? if so why also pass in an object instance? Is this some sort of combined GetBlockByName and GetBlockByObjectInstance? If so these should be separate functions.

JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger.c
```diff
@@ -0,0 +1,199 @@
...
+
+	dbg_Helicopters.Add("Position", GetPosition());
+	dbg_Helicopters.Add("Orientation", GetOrientation());
+
+	CF.Debugger.Display("Helicopters", this, "Velocity", GetVelocity(this));
```
Arkensor: Where does the `Display` method come from? What do these parameters mean?

JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger_Block.c
```diff
@@ -0,0 +1,144 @@
...
+
+	private Object m_Target;
+	private bool m_HasTarget;
+
+	private bool m_Buffered;
+	private bool m_SwapClears;
+
+	private int m_Count;
+
+	private autoptr map<string, int> m_EntryMap;
```
Arkensor: Why an array index map to access the values instead of just a string, value map?

JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger_Block.c
```diff
@@ -0,0 +1,144 @@
...
+};
+
+class CF_Debugger_Block: ScriptViewTemplate<CF_Debugger_Block_Controller>
+{
+	private string m_Name;
+
+	private Object m_Target;
+	private bool m_HasTarget;
+
+	private bool m_Buffered;
```
Arkensor: What does buffering mean in this context? What is buffered and why? When is it flushed?

JM/CF/Debugger/layouts/Debugger_Block.layout
```diff
@@ -0,0 +1,57 @@
...
+  MultilineTextWidgetClass EntryValues {
+   ignorepointer 1
+   position 0.4 0
+   size 0.6 0.98
+   hexactpos 0
+   vexactpos 0
+   hexactsize 0
+   vexactsize 0
+   scriptclass "ViewBinding"
+   text "deserunt mollit anim id est laborum.\""\
```
Arkensor: Can we remove this sample text for the release and just replace it with a short "PLACEHOLDER"?

JM/CF/Scripts/3_Game/CommunityFramework/CommunityFramework.c
```diff
@@ -42,6 +43,8 @@ class CommunityFramework
 		#ifdef CF_MODULE_PERMISSIONS
 		CF_Permission_ManagerBase._Init( Permission );
 		#endif
+        
+        Debugger = new CF_Debugger();
```
Arkensor: Why is this on world init and not game init? Is it because of UI components?

JM/CF/Debugger/3_Game/CF_Debugger/CF_Debugger.c
```diff
@@ -0,0 +1,199 @@
...
+}
+
+*/
+
+class CF_Debugger
+{
+	private autoptr array<ref CF_Debugger_Block> m_Blocks;
```
Arkensor: I think we should use protected. I see no reason why someone could not inherit their own variant and replace their local CF.Debugger instance at runtime with their own type.

JM/CF/Debugger/3_Game/CF_Debugger/GUI/CF_Debugger_Block_Controller.c
```diff
@@ -0,0 +1,5 @@
+class CF_Debugger_Block_Controller: Controller
+{
+	string EntryKeys;
```
Arkensor: Keys = plural? How are multiple keys stored here?

JM/CF/Scripts/3_Game/CommunityFramework/CommunityFramework.c
```diff
@@ -51,6 +54,8 @@ class CommunityFramework
      */
     static void _Cleanup()
     {
+        delete Debugger;
```
Arkensor: No cleanup required? To safely close UI and delete all data?

Arkensor (changes requested):
With the current intention of this I suggest the module name `DebugUI` instead of `Debugger` to avoid any confusion on what this is for and how it can be used.

The fully qualified name would be `CF.DebugUI / CF_DebugUI`

Also, where is the logic that opens the GUI and maintains data updates?

Issue #47, opened 2021-04-08
ModStorage hard crashes when loading storage


lava76:
It's due to a change in this commit: 25f7b202bc2f88e91e7307a347b72c9e7e536d67

Reverting the commit fixes the problem.

PR #55, opened 2021-06-07
Save the hash of the mod name instead of the mod name itself

lava76:
This considerably reduces the amount of data written to/read from storage if a number of mods use modstorage.
Compatibility with the old way of using the mod name is preserved and there are no changes needed in CF_OnStoreLoad/Save.


Arkensor:
Hello @lava76,

thank you for your PR. @Jacob-Mango should review this as he is the maintainer of modstorage, but the general idea seems to be a good addition from my point of view.

Semi-unrelated:
Why did you add all the refs? Was there something losing the reference and resulting in an error? If so can you please provide a quick example of that?

lava76:
> Why did you add all the refs? Was there something losing the reference and resulting in an error?

It was to fix this hard crash (https://github.com/Jacob-Mango/DayZ-Community-Framework/issues/47) which was a result of some of the changes in [25f7b20](https://github.com/Jacob-Mango/DayZ-Community-Framework/commit/25f7b202bc2f88e91e7307a347b72c9e7e536d67).

Arkensor:
A potential problem @Jacob-Mango and I discussed is the probability of a mod name hash collision. The function to calculate the hash is the following:
```
uint hash = 0
foreach(character in string)
    hash = hash * 37 + ASCIIValue(character)
return hash
```
Have you considered this problem yet?

lava76:
Thought about it, although I didn't test for how likely a collision would be. Worst come to worst, I could roll my own hash function as it's only meant to be called once for each mod at server start, so shouldn't have much of a performance impact.

Arkensor:
A collision check was a good addition either way.

lava76:
I did a bit of testing. Using this list of 216,553 english words: https://drive.google.com/file/d/0B3BLwu7Vb2U-dEw1VkUxc3U4SG8/view?usp=sharing
DayZ Hash() function seems to be pretty good for actual word-like character strings (no collisions in the above word list, unlike some of the more common 32-bit hashing functions like crc32 with 6 collisions or fnv1a-32 with 9), not so much for strings with a lot of character repetition which resulted in abysmal collisions (tested using combinations of 0-9, A-Z, a-z and _), which is to some extent also true for crc32 and fnv1a-32.

lava76:
> DayZ Hash() function seems to be pretty good for actual word-like character strings (no collisions in the above word list, unlike some of the more common 32-bit hashing functions like crc32 with 6 collisions or fnv1a-32 with 9)

I made a mistake: I accidentally tested with a 64-bit version of the hash function (that's what I get for prototyping in Python). The 32-bit version produces 7 collisions, so roundabout the same amount as the other two. Also, forgot to mention that I used capitalized versions of the words, not all uppercase, for the tests.
If I change the 37 in the original function to 109, that reduces collisions to zero for the word list and the 16,007,040 one-to-four-letter-combinatons of 0-9, A-Z, a-z and _ that I tested, but it still collides (after generating roughly 134,758 hashes without collisions) when using combinations of words.
If I use two hashes instead of one (2nd one uses the string to be hashed in reverse order) I couldn't produce a collision even with the original hash function in 100,000,000 combinations of the words (I aborted there), so that could be a way to make it more robust.

Arkensor:
Hello @lava76,
we will continue with PR in the near future when the work on an RC for modstorage is picked up again (most likely after the next DayZ patch).

PR #57, opened 2021-07-07
Removed MVC

Jacob-Mango:


InclementDab:
Branch is compatible with DayZ 1.12 and 1.13, along with DF and Editor on both versions.

Arkensor (approved):

PR #58, opened 2021-07-11
Scripted streams

Jacob-Mango:


JM/CF/Scripts/3_Game/CommunityFramework/Stream/CF_BinaryStream.c
```diff
@@ -0,0 +1,286 @@
...
+	override void WriteCString(string value)
+	{
+		for (int i = 0; i < value.Length(); i++)
+		{
+			Write(value[i].Hash());
+		}
+		Write(0);
+	}
+
+	override int Read()
```
Arkensor: Read ... what? A byte, if so this should be ReadByte()

Otherwise causes confusion with the other ReadX methods.

JM/CF/Scripts/1_Core/CommunityFramework/CF_Cast.c
```diff
@@ -0,0 +1,13 @@
...
+	{
+		U src_arr[1];
+		src_arr[0] = src;
+		V dst_arr[1];
+
+		copyarray(dst_arr, src_arr);
+
+		return dst_arr[0];
+	}
+};
```
Arkensor, line 13: New line at the end of files please.

JM/CF/Scripts/3_Game/CommunityFramework/Stream/CF_BinaryStream.c
```diff
@@ -0,0 +1,286 @@
...
+
+		int index;
+
+		switch (mode)
+		{
+			case FileMode.READ:
+			{
+				int lastZero = 0;
+
+				bool varr[1];
```
Arkensor: More descriptive name for `varr`

JM/CF/Scripts/3_Game/CommunityFramework/Stream/CF_BinaryStream.c
```diff
@@ -0,0 +1,286 @@
...
+	{
+		return m_Position;
+	}
+
+	override int Length()
+	{
+		return m_Data.Count();
+	}
+
+	override void GoTo(int position)
```
Arkensor: Instead of GoTo Use Seek with an origin enum. 

s. below

JM/CF/Scripts/3_Game/CommunityFramework/Stream/CF_BinaryStream.c
```diff
@@ -0,0 +1,286 @@
...
+	{
+		m_Position = position;
+	}
+
+	override void Seek(int offset)
```
Arkensor: Prototype should be `void Seek(int offset, CF_Stream_Origin origin)` where origin is an Enum with for example `SEEK_SET`, `SEEK_CUR`, `SEEK_END` ... inspired by [c++ fseek](https://www.cplusplus.com/reference/cstdio/fseek/)

Default value for origin should be `SEEK_CUR`

JM/CF/Scripts/3_Game/CommunityFramework/Stream/CF_Stream.c
```diff
@@ -0,0 +1,183 @@
...
+	{
+		/*CF.Log.*/Error("Not implemented.");
+	}
+
+	void Seek(int offset)
+	{
+		/*CF.Log.*/Error("Not implemented.");
+	}
+
+	void SetBytes(array<int> bytes)
```
Arkensor: Why is there a SetBytes method? Normally streams are only interacted with via Read and Write and the only full data operation is GetBytes?

JM/CF/Scripts/3_Game/CommunityFramework/XML/CF_XML_Callback.c
```diff
@@ -0,0 +1,10 @@
+typedef CF_XML_Callback XMLCallback;
```
Arkensor, line 1: Not sure why git did not detect this as a file location move. Please validate that there were no accidental changes.

JM/CF/Scripts/3_Game/CommunityFramework/XML/CF_XML.c
```diff
@@ -4,7 +4,7 @@ class CF_XML
 	{
 	}
 
-	static ref CF_XML_Reader Open(string path)
```
Arkensor, line 7: Are you 100% sure that **ALL** the refs should be removed? (In this file and the other XML related ones)
If so good, otherwise please check if they were needed somewhere.

Arkensor (changes requested):
A few minor questions and suggestions. Please check all of them before requesting a re-review.

JM/CF/Scripts/1_Core/CommunityFramework/CF_Cast.c
Jacob-Mango: Weird, it has a new line for me on my PC. Git isn't detecting new file changes.

JM/CF/Scripts/3_Game/CommunityFramework/Stream/CF_Stream.c
Jacob-Mango: Syncing over the network through RPC (although inefficient).

Generally, people shouldn't use it but giving options is better than nothing

JM/CF/Scripts/1_Core/CommunityFramework/CF_Cast.c
Arkensor: I do not think git is to blame here, even the addition of an empty line would be a stageable change

Arkensor: Only Github shows it like this ... when cloned there is a newline.

Arkensor (approved):

docs/Streams/index.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+### **Saving a File**
+
+```
+CF_BinaryStream stream();
+if (stream.File("$profile:Test.png"))
+{
+  array<int> data = stream.GetBytes();
+  GetRPCManager().
```
Arkensor: I'd leave RPC out of this tbh. Also, this line is incomplete regardless.

docs/Streams/index.md
```diff
@@ -0,0 +1,83 @@
...
+**text.txt**:
+```
+Hello, World!
+```
+
+#### Read
+
+```
+CF_StringStream stream();
+stream.File("$profile:test.txt", FileMode.READ);
```
Arkensor: Maybe have a look at c++ fstream. It had an optional constructor parameter where you could put in the file and the mode already which internally called Open

I think binarystream and stringstream are ok, but if you want to interact with files, then that should be a CF_Filestream ... which you could open from one of the other streams to get their data into the file or read from the file again. E.g.

```cs
CF_StringStream data();
data.Write("Hello Jacob");
CF_FileStream fileWrite("<path>", Filemode.OPTION);
fileWrite.Write(data);
fileWrite.Close();
```

Your examples all show file interaction ... maybe you should first demonstrate the streams for what they are named after, aka reading and writing strings or binary data ... and the use-case together with files should be coming a bit later - or be its own kind of stream.

Arkensor (commented):
Conceptual suggestions

JM/CF/Scripts/1_Core/CommunityFramework/CF_Operations.c
```diff
@@ -0,0 +1,8 @@
+static int ^(int x, int y)
+{
+	int a = (x | y);
+	int b = ~x;
+	int c = ~y;
+	int d = (b | c);
+	return (a & d);
+}
```
Arkensor: Missing newline

docs/Streams/index.md
```diff
@@ -8,76 +8,79 @@ There are currently two Streams avaliable, `CF_BinaryStream` and `CF_StringStrea
...
-array<int> data = // ...
+```csharp
+array<CF_Byte> data = new array<CF_Byte>();
 
-CF_BinaryStream stream();
+CF_BinaryStream stream = new CF_BinaryStream();
 if (stream.File("$profile:Test.png", FileMode.WRITE))
 {
-   stream.SetBytes(data);
+  stream.SetBytes(data);
```
Arkensor: Wrong indent

JM/CF/Scripts/1_Core/CommunityFramework/CF_Byte.c
```diff
@@ -0,0 +1,73 @@
...
+*/
+
+typedef int CF_Byte;
+class CF_Byte : CF_Uint
+{
+	CF_Byte Clamp()
+	{
+		CF_Byte old_byte = value;
+		value = value & 255;
+		if (value != old_byte)
```
Arkensor: Remove snake case.

JM/CF/Scripts/1_Core/CommunityFramework/CF_Byte.c
```diff
@@ -0,0 +1,73 @@
...
+
+		return CF_Encoding.BASE_16[n0] + CF_Encoding.BASE_16[n1];
+	}
+
+	override CF_Uint ^(CF_Uint y)
+	{
+		CF_Byte b = super.^(y);
+		b.Clamp();
+		return b;
+	}
```
Arkensor: Are we really keeping this? I know it is cool at a first glance but in code ehhh ... I think XOR written out below is enough.

JM/CF/Scripts/1_Core/CommunityFramework/CF_PackedByte.c
```diff
@@ -0,0 +1,117 @@
+class CF_PackedByte
```
Arkensor, line 5: What is a packed byte? Does this mean bit stuffing as in multiple values in one byte?

JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_Base16Stream.c
```diff
@@ -0,0 +1,110 @@
...
+		m_Position = oldPosition;
+		m_Current = oldCurrent;
+
+		m_Dirty = false;
+	}
+
+	override string ToStr()
+	{
+		return super.ToStr();
+		//return Encode();
```
Arkensor: remove unused code

JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_IO.c
```diff
@@ -0,0 +1,208 @@
+class CF_IO : Managed
```
Arkensor, line 1: I understand the idea of a base stream, and it is good, but I would add one abstraction layer in between, to make sure you do not come to the idea to Read on a write stream or vice-versa.
Just CF_IO should not allow to read and write, one has to decide what they want it to be by strong typing it to something higher.

Move WriteX into inherited CF_WriteStream
Move ReadX into a CF_ReadStream

And inherit the other readers and writers from those intermediate classes.

docs/Streams/index.md
```diff
@@ -0,0 +1,73 @@
+# Streams
+
+A stream is a doubly linked list of bytes. It can be used for file input/output, DayZ serializers, hex conversion or string parsing. Once a stream is finished, it is recommended to close the stream either by letting the reference count equal zero or call the `Close` method.
```
Arkensor: Once you do not longer need the stream, it is...

docs/Streams/index.md
```diff
@@ -0,0 +1,73 @@
...
+// Write to the context
+CF_BinaryWriter writer = new CF_BinaryWriter(new CF_SerializerWriteStream(ctx.GetWriteContext()));
+writer.WriteFloat(0.1);
+writer.Close();
+
+// Read the context
+CF_BinaryReader reader = new CF_BinaryReader(new CF_SerializerReadStream(ctx.GetReadContext()));
+Print(reader.ReadFloat());	// "0.1"
+reader.Close();
+```
```
Arkensor, line 80: Missing example for SHA256 ...

docs/Streams/index.md
```diff
@@ -0,0 +1,73 @@
...
+Streams can be manipulated with `CF_TextWriter`, and `CF_BinaryWriter`. Streams can be read with `CF_TextReader`, and`CF_BinaryReader`.
+
+## Files
+
+Use the `CF_FileStream` class to read and write files on the a file system. 
+
+```csharp
+string path = "$profile:test.txt";
+
+if (FileExist(path)) DeleteFile(path);
```
Arkensor: The file mode should determine if the file is replaced or appended, so this line has no real value does it? I would remove it.

docs/Streams/index.md
```diff
@@ -0,0 +1,73 @@
...
+Use the `CF_HexStream` class to read and write to a string source as a hex output.
+
+```csharp
+CF_HexStream stream = new CF_HexStream();
+
+CF_BinaryWriter writer = new CF_BinaryWriter(stream);
+writer.WriteFloat(5);
+writer.Close();
+
+Print(stream.ToStr());	// "40A00000"
```
Arkensor, line 58: If you make an example about float, maybe append the endianness and float format to the comment.

Arkensor (changes requested):
Overall very solid, if it is all tested and works as intended then this will be a great feature. Only a few change requests, the rest is questions or suggestions.

I thought about if you could add a `string CF_Encoding.ToHex(int value)` and `int CF_Encoding.FromHex(string value)` shortcuts so one must not go through the steam overhead.

JM/CF/Scripts/1_Core/CommunityFramework/CF_Byte.c
Jacob-Mango: Yes

JM/CF/Scripts/1_Core/CommunityFramework/CF_PackedByte.c
Jacob-Mango: Linked list of `CF_Byte`, optimized reading and writing.

JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_IO.c
Jacob-Mango: Doesn't solve anything by having an additional layer of abstraction, just makes it harder to re-use the same variable name in the same scope.

docs/Streams/index.md
Jacob-Mango: I don't understand your statement

JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_IO.c
Arkensor: This solves not knowing what interaction is allowed on a given stream. You are not supposed to first use a stream instance to read and then turn it into a writer. Those are usually two separate streams, of two separate types. Reading only on a ReadStream and Writing only on a WriteStream.

docs/Streams/index.md
Arkensor: Your wording just sounds odd. Replace "Once a stream is finished" with what I said/rewrite it. A stream is not finished - it is not being used anymore or empty.

JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_IO.c
Jacob-Mango: I will add a return method for `IsRead` and `IsWrite`. 

There may be a use case in the future where an InputOutput operation can be done at the same time. Rather not lock CF into a forced inheritance.

Jacob-Mango:
> I thought about if you could add a string CF_Encoding.ToHex(int value) and int CF_Encoding.FromHex(string value) shortcuts so one must not go through the steam overhead.

This already exists in a slightly different syntax.

To Hex:

```cpp
CF_Byte value = 80;
string hex = value.ToHex();
```

```cpp
array<CF_Byte> bytes = new array<CF_Byte>();
bytes.Insert(80);
string hex = CF_Encoding.ToHex(bytes);
```

From Hex:

```cpp
string hex = "FF";
array<CF_Byte> bytes = CF_Encoding.FromHex(hex);
```

JM/CF/Scripts/1_Core/CommunityFramework/CF_Byte.c
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+typedef int CF_Byte;
+class CF_Byte : CF_Uint
+{
+	CF_Byte Clamp()
+	{
+		CF_Byte oldByte = value;
+		value = value & 255;
+		return value;
```
Arkensor, line 17: Why not 
```
return (value & 255);
```
to save the assignment

Or is this about in-place modification and returning the value is just for convinience?

JM/CF/Scripts/1_Core/CommunityFramework/CF_Byte.c
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	void Set(int index, int v)
+	{
+		value = v;
+		value.Clamp();
```
Arkensor: It might be significantly faster to not make Clamp a function and instead have `& 255` in the code each time.

JM/CF/Scripts/1_Core/CommunityFramework/CF_Long.c
```diff
@@ -0,0 +1,12 @@
...
+{
+	int I0;
+	int I1;
+
+	string GetDebugName()
+	{
+		return "" + I0 + I1;
+	}
+};
+*/
```
Arkensor: Unused or discarded concept? Then delete the file, please.

JM/CF/Scripts/1_Core/CommunityFramework/CF_Uint.c
```diff
@@ -0,0 +1,138 @@
...
+
+typedef int CF_Uint;
+class CF_Uint : int
+{
+	CF_Uint Add(CF_Uint b)
+	{
+		CF_Uint a = value;
+		CF_Uint carry = a & b;
+		CF_Uint result = a.XOR(b);
+		while (carry != 0)
```
Arkensor: You can just do `while(carry)` because any value other than 0 is implicitly true. Saves the comparison on each iteration.

docs/Cryptography/index.md
```diff
@@ -0,0 +1,21 @@
...
+
+A variety of cryptography functions will be added to Community Framework in the future. 
+
+## SHA-256
+
+```csharp
+string path = "$profile:test.txt";
+
+// Create the file
+CF_TextReader reader = new CF_TextReader(new CF_FileStream(path, FileMode.READ));
```
Arkensor: rename reader to `input`

docs/Cryptography/index.md
```diff
@@ -0,0 +1,21 @@
...
+Print(reader.ReadLine()); // "This is some text"
+
+CF_Base16Stream hexStream = new CF_Base16Stream();
```
Arkensor: rename hexStream to `output`

That makes it easier to understand what is what in the `Process` call

docs/Streams/index.md
```diff
@@ -0,0 +1,73 @@
...
+string data = "Hello, World!";
+
+CF_TextReader reader = new CF_TextReader(new CF_StringStream(data));
+Print("ReadWord: " + reader.ReadWord());				// "Hello"
+Print("ReadChar: " + reader.ReadChar());				// ","
+Print("ReadWhitespace: " + reader.ReadWhitespace());	// ""
+Print("ReadWord: " + reader.ReadWord());				// "World"
+Print("ReadChar: " + reader.ReadChar());				// "!"
+reader.Close();
+```
```
Arkensor, line 43: Maybe you could give a bit of information on how we detect what a word is? I would assume it reads until the next separator char? If so a quick list of which chars for this would be nice. Or if it's the rule "characters that are not aA-zZ" then say that.

docs/Streams/index.md
```diff
@@ -0,0 +1,73 @@
...
+CF_BinaryWriter writer = new CF_BinaryWriter(stream);
+writer.WriteFloat(5);
+writer.Close();
+
+Print(stream.ToStr());	// "40A00000"
+```
+
+## Serializer
+
+Use the `CF_SerializerReadStream` class to read to from the DayZ serializer and `CF_SerializerWriteStream` to write to the DayZ serializer.
```
Arkensor: I would clarify that it can interact with anything that inherits from `Serializer`, otherwise one could think you only mean ScriptReadWriteContext. 

> Use `CF_SerializerReadStream`/`CF_SerializerWriteStream` for read and write interaction with anything inherited from the DayZ `Serializer` class.

Arkensor (changes requested):
Just a few minor comments to double-check before we can merge this.

JM/CF/Scripts/1_Core/CommunityFramework/CF_Byte.c
Jacob-Mango: > Or is this about in-place modification and returning the value is just for convenience?

Correct, so both `cfbyte.Clamp()` and `x = cfbyte.Clamp()` compile

Arkensor (approved):

PR #70, opened 2021-08-06
Initial implementation of timers

Jacob-Mango:


docs/Timers/index.md
```diff
@@ -0,0 +1,71 @@
+# Timers
+
+## Instance Timers
+
+Memory management is handled all for you. To create a timer you run `CF_Timer.Create` with the arguments you specify. Note that an interval of `0.0` means the timer will tick every frame.
```
Arkensor: Are you sure it's every frame and not every **simulation step**? If it is OnFrame, then I recon to change to OnSimulation, since only those progress states. Every frame would most likely execute code multiple times with any effect for the person. There can be multiple OnFrame events between one OnSimulate/OnUpdate

JM/CF/Scripts/3_Game/CommunityFramework/Timer/DayZGame.c
```diff
@@ -0,0 +1,9 @@
+modded class DayZGame
+{
+	override void OnUpdate(bool doSim, float timeslice)
+	{
+		super.OnUpdate(doSim, timeslice);
+		
+		CF_TimerBase.Update(timeslice);
```
Arkensor: s. Comment on docs about OnFrame vs OnSimulate. This might need to be changed to 
```cs
if (doSim) CF_TimerBase.Update(timeslice);
```

docs/Timers/index.md
```diff
@@ -0,0 +1,71 @@
...
+
+```csharp
+class TestClass : Managed
```
Arkensor: Does managed matter here? If not I would remove it, otherwise it implies that timers only work on managed wrappers.

docs/Timers/index.md
```diff
@@ -0,0 +1,71 @@
...
+{	
+    void TestClass()
+    {
+		Widget widget = ...
+		float timeToFade = ...
+        CF_Timer.Create(this, "FadeIn", 0.0, new CF_TimerParam1<Widget, float>(widget, timeToFade));
```
Arkensor: The parameters should be explained by showing the function prototype with their default values in the text above when referring to `CF_Timer.Create`

Also are you sure it's "FadeIn" and not "TimerFunction"? Because otherwise, I do not understand the connection between the function and the timer at all.

docs/Timers/index.md
```diff
@@ -0,0 +1,71 @@
...
+    {
+        if (timer.GetTime() > timeToFade * 1000.0)
+        {
+            timer.Stop();
+            return;
+        }
+
+		float alpha = timer.GetTime() / (1000.0 * timeToFade);
+        widget.SetAlpha(Easing.EaseInCubic(alpha));
+    }
```
Arkensor: That example is way too much. You want to explain the basic functionality of a timer, so please put in the most basic possible way to setting up and using a timer ... with something like `Print("Tick")` or whatever in the function body. You can add some advanced examples further into the docs with some tips and tricks if you think people would miss out on something otherwise.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
+class CF_TimerBase : Managed
+{
+	private static ref CF_TimerBase s_Root;
```
Arkensor: `s_` for static? I'd call it `s_Tail` or `s_Head` depending if the list is append or prepend

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+	protected void CF_TimerBase()
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "CF_TimerBase");
+		#endif
+
+		OnCreate();
+	}
+
+	void ~CF_TimerBase()
```
Arkensor, line 24: protected constructor but public destructor? Intentional?

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+
+		if (s_Root == null)
+		{
+			s_Root = this;
+		}
+		else
+		{
+			m_Next = s_Root;
+			s_Root.m_Prev = this;
+			s_Root = this;
```
Arkensor: I am a bit confused with the order here. You seem to build the list from right to left. An insert gives root a previous and the current ones next is the root, and then root shifts to the current one.

Given that we have one element 0, this means on insert we create -1 and not 1 ... seems kinda odd to do this that way around. Why not store s_Head and s_Tail and always append on s_Tail with 
```cs
s_Tail.m_Next = this;
this.m_Prev = s_Tail;
s_Tail = this;
```

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	void SetInterval(float duration)
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "SetInterval", "" + duration);
+		#endif
+
+		m_Interval = duration;
+		if (m_Interval < 0) m_Interval = 0;
```
Arkensor: Change to
```cs
if(duration >= 0) m_Interval = duration;
```

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+{
+	private static ref CF_TimerBase s_Root;
+	private ref CF_TimerBase m_Next;
+	private CF_TimerBase m_Prev;
+	private bool m_IsActive;
+	private bool m_Destroy;
+
+	protected float m_Interval; // Seconds
+	protected float m_DeltaTime; // Seconds
+	protected int m_Time; // Milliseconds
```
Arkensor: I think it should all be MS or all be seconds. It is confusing especially since these variables are exposed through getter and setter. The original timer uses seconds so I'd go for that.

Arkensor: More comments on `m_Time` at its setter/getter

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	int GetTime()
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "GetTime");
+		#endif
+
+		return m_Time;
+	}
```
Arkensor: What even is time ... *thoughtful music plays*

I think a more descriptive name would be `m_TimeElapsed` (borrowed from C# TimeSpan) and it should be also in float seconds if the rest is in seconds. The original DayZ timer also used seconds!

Function could be `SetTimeElaspsed(float elapsed)` and `float GetTimeElapsed()`

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+
+	void SetTime(int time)
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "SetTime", "" + time);
+		#endif
+
+		m_Time = time;
+	}
```
Arkensor: Setting the elapsed time ... why? Resetting a timer sure, but then I would only add a Reset() function tbh ... setter feels odd here. It implies that setting values other than 0 are intentional by design, which I have not been able to come up with a use-case for.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+
+		return m_Time;
+	}
+
+	protected void OnCreate()
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "OnCreate");
+		#endif
+
```
Arkensor: Remove empty line.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	protected void OnDestroy()
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "OnDestroy");
+		#endif
+
```
Arkensor, line 128: Remove empty line.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	protected void OnStart()
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "OnStart");
+		#endif
+
```
Arkensor, line 132: Remove empty line.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	protected void OnStop()
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "OnStop");
+		#endif
+
```
Arkensor: Remove empty line.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	protected void OnTick(float dt)
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "OnTick", "" + dt);
+		#endif
+
```
Arkensor, line 146: Remove empty line.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+		m_Time += dt * 1000.0;
+		m_DeltaTime += dt;
+
+		if (m_DeltaTime >= m_Interval)
+		{
+			OnTick(m_DeltaTime);
+			m_DeltaTime = 0;
+		}
```
Arkensor: Not an issue but a suggestion:

The code design does not take into account a lag spike. You can ignore them, but then OnTick will have a different delta-time quite often. A fixed timestep there might be more comfortable.

```cs
while(m_DeltaTime >= m_Interval)
{
	m_DeltaTime -= m_Interval;
	OnTick(m_Interval);
}
```

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+
+	static void Update(float dt)
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(null, "CF_TimerBase::Update");
+		#endif
+
+		CF_TimerBase current = s_Root;
+		CF_TimerBase next = null;
+		while (current != null)
```
Arkensor: change to 
```cs
while (current)
```
Yes I know != null is more readable, but this is faster and since it's a non-public function and is often called we should optimize every instruction out of it.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+		{
+			next = current.m_Next;
+			current.OnUpdate(dt);
+			if (current.m_Destroy) delete current;
```
Arkensor: The only thing that keeps a timer alive is the strong ref to the m_Next. So if you delete current how do you make sure m_Next is still allowed to live? If a timer is deleted in the middle of the linked list this fails unless I am missing something atm.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,202 @@
...
+		m_DeltaTime += dt;
+
+		if (m_DeltaTime >= m_Interval)
+		{
+			OnTick(m_DeltaTime);
+			m_DeltaTime = 0;
+		}
+	}
+
+	static void Update(float dt)
```
Arkensor: Since this is an internal method for the abstracted-away functionality I would say it's `_Update(float dt)` since users shall never call this themselves.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_Timer.c
```diff
@@ -0,0 +1,55 @@
+class CF_Timer : CF_TimerBase
+{
+	protected Managed m_Object;
```
Arkensor: A better name could be `m_Instance`... `m_Object` would imply a world object created by `GetGame().CreateObject()`

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_Timer.c
```diff
@@ -0,0 +1,55 @@
...
+	static CF_TimerBase Create(Managed obj, string funcName, float duration = 0.025, CF_TimerParam params = null)
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(null, "Create", "" + obj, "" + funcName, "" + duration, "" + params);
+		#endif
+
+		if (obj == null) return null;
+
+		CF_Timer timer = new CF_Timer();
+		timer.m_Object = obj;
```
Arkensor: s. comment above about m_Object. The parameter name needs to be adjusted accordingly.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_Timer.c
```diff
@@ -0,0 +1,55 @@
...
+		timer.m_Function = funcName;
+		timer.m_Params = params;
+		if (!timer.m_Params) timer.m_Params = new CF_TimerParam();
```
Arkensor: A bit cleaner imo
```cs
if(!params) params = new CF_TimerParam();
timer.m_Params = params;
```

Why do we assign empty new params? Can't we just invoke the timer function with null? If people did not care about having them in their Create method then they could omit it from their function they subscribe for the OnTick and not care about it. Also saves on performance.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerParam.c
```diff
@@ -0,0 +1,167 @@
+class CF_TimerParam
+{
+	CF_TimerBase m_Timer;
```
Arkensor, line 10: `m_Timer` is never assigned.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_Timer.c
```diff
@@ -0,0 +1,55 @@
...
+{
+	protected Managed m_Object;
+	protected string m_Function;
+	protected ref CF_TimerParam m_Params;
+	
+	void CF_Timer()
+	{
+	}
+
+	static CF_TimerBase Create(Managed obj, string funcName, float duration = 0.025, CF_TimerParam params = null)
```
Arkensor: Change ` float duration = 0.025` to ` float interval = 0.0`. Besides the name I would say that people should actively choose to only execute on 25 fps etc, the default - if given - should be every tick.
OR if you want that people think about it, do not make this param optional - that would be a good change regardless in my opinion.

Arkensor (changes requested):
Alright, lots of thoughts on this. The docs need some more general and easier examples, also there is no real guideline to tell when to choose CF_Timer and CF_TimerBase. I still don't understand the split after having read all of the code.

What I understood is that CF_Timer is essentially CF_TimerBase but you pass in the wrapper instance, function to call on each tick and the tick interval. And you can pass some data in there that is stored and passed into the tick invoke each time. That is all useful, but I don't quite get any anyone would want to NOT use that and JUST use CF_TimerBase.

docs/Timers/index.md
Jacob-Mango: Yes, I do mean every frame. For example, moving a panel across the screen. It would look and feel smoother when done per frame at an unlocked frame rate.

docs/Timers/index.md
Jacob-Mango: CF_Timer does only work on Managed classes

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: Intentional so no one creates an instance of `CF_TimerBase`. They must sub-class or use `CF_Timer`

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: Storing `m_Time` in seconds will lead to floating-point errors when the timer/server has been alive for more than 2.5 hours. I would rather store it all in seconds but I'm keeping that in mind.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: > which I have not been able to come up with a use-case for.

Syncing the world event time from the server to newly connected clients

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: All variables are strong references inside functions.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_Timer.c
Jacob-Mango: By default, that means it executes every 40hz. Which is the same rate world Objects simulate at.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: Intentional, to make sure m_Interval never falls below `0 seconds`.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_Timer.c
Jacob-Mango: The function call contains `CF_TimerBase,float` before the rest of the params continues. That is why.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: Passing in delta time handles that already, on the off chance a modders `OnTick` is the cause of the lag spike, it would prevent a worse situation.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: Requires more checking on inserting/removing and serves no benefit to store a head and tail.

Do you see a reason why we should be adding to the tail, aside from convention?

Jacob-Mango:
> but I don't quite get any anyone would want to NOT use that and JUST use CF_TimerBase.

There would be minimal use-cases when one would choose `CF_Timer` over `CF_TimerBase`, yes. When I asked other modders, what I got in reply was that it is a good substitute for `ScriptCallQueue::CallLater` since that function degrades after 6 hours of server up-time.

docs/Timers/index.md
Arkensor: Why would it matter where a CF_Timer is contained in? I refer to `TestClass`

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Arkensor: No, I mean why is the destructor public? If the constructor is protected, the destructor might as well also be.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Arkensor: How do you iterate from the start? Cause your s_Root keeps shifting, so you lose the info on what the first list item was? The code is so confusing,  I can't really follow it.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Arkensor: Ok, storing time as ms in an int variable is enough for 25 days of runtime. Since we have the floating-point issue with all of the member vars, I would suggest we change all 3 to int so people do not need to look up what is *1000 or /1000 ... It is very common for ms intervals to be passed as ints and not floats in other languages, so I think this would be no issue.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Arkensor: Though this sync needs some extra care by transmitting simulation steps. If you just send the ms count via RPC it will be out of sync by the time it arrives. *Maybe* this problem is worth providing a built-in solution as a wrapper on top or static functionality of a `CF_NetTimer` `CF_SyncedTimer` ... something like that.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: There is no need to keep the root consistent, nor is there a need to make sure a timer is updated at the insert position when iterating.

The iterating starts at the most recently inserted Timer, and the last one updated is the oldest running timer.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: I can make it so member variaes are ints and milliseconds. 

But the arguments passed into the OnUpdate and OnTick methods will continue to remain floats and be in seconds, to keep consistent with every other Update method within vanilla  DayZ scripts

Arkensor: The problem is that it is super ugly to have float parameters to pass in but in the function logic, you suddenly need to divide or multiply by 1000 because the member vars you interact with stored them this way. So no, I think it should be all ints.

Jacob-Mango: No valid use-case or equation works with milliseconds. They all work seconds.

Will just rename the `SetTimeElapsed` to `SetTimeElapsedMS` and `GetTimeElapsed` to `GetTimeElapsedMS`, keeping everything else as floating point numbers.

Arkensor: What about we internally use an int to store it as m_ElapsedMS and also provide m_Elasped as float in seconds, that way people can choose what floats the boat. And the parameters are all gonna be in float.
We just update the float with `m_Elasped = m_ElapsedMS / 1000.0;`

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_DelayCall.c
```diff
@@ -0,0 +1,69 @@
...
+{
+	protected Managed m_Instance;
+	protected string m_Function;
+	protected ref CF_TimerParam m_Params;
+
+	protected int m_Delay;
+	
+	void CF_DelayCall()
+	{
+	}
```
Arkensor: Why is this here? If we only want them to work via CF_DelayCall.Create(), then ctor and dtor should be protected or private.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_DelayCall.c
```diff
@@ -0,0 +1,69 @@
+class CF_DelayCall : CF_TimerBase
```
Arkensor: 1. I am not sure what the vision is for DelayCall ... the name suggests delaying ONE function call to a later time ... aka the same that `ScriptCallQueue` already does.
Do you maybe mean `RepatedCall`/`ReapeatableCall` since we need to store the result of `CF_DelayCall::Create()` somewhere already? 
If it was just for a one-time call the storage should be static and the return value `void` - e.g. 
```cs
void main()
{
	CF_DelayCall.Create(inst, "SomeFunc", 0.5, ...)
}
```

2. If the "delay" bit stays: `CF_DelayedCall` or `CF_DelayCall` since we will interact with this class only via `Create` .. and it sounds nicer to create a *delay**ed*** call instead of a delay call IMO.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_DelayCall.c
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * @param instance The instance that owns the function
+	 * @param function The function that is called
+	 * @param interval The time between each function call, in milliseconds.
+	 * @param params The static parameters that are passed into the function
+	 * 
+	 * @note The function passed must have a signature that starts with 'CF_TimerBase,float` and then continues with the matching 'params'.
+	 */
+	static CF_TimerBase Create(Managed instance, string function, int delay, CF_TimerParam params = null)
```
Arkensor: 1. Does it **need** managed here? Same discussion we had about normal timers - if we are responsible to handle the case when the instance that we should invoke becomes invalid.
Changing it to class would allow this to be used with existing custom classes that use inheritance. 

2. Change the delay param to be float to fit the rest of the timer functions.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_ManagedTimer.c
```diff
@@ -0,0 +1,20 @@
+class CF_ManagedTimer : CF_Timer
+{
+	Managed m_ManagedInstance;
+	
+	void CF_ManagedTimer(Class instance, string function, float interval = 0.0, CF_TimerParam params = null)
```
Arkensor: This should be `Managed` and not` Class` because one might create a managed timer outside of the timer factory. You already pass it in via `Managed` in the factory code.

Personal note: I prefer NULL over null. Calls more attention to itself.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_Timer.c
```diff
@@ -0,0 +1,69 @@
+class CF_Timer : CF_TimerBase
+{
+	Class m_Instance;
+	string m_Function;
+	ref CF_TimerParam m_Params;
```
Arkensor: Do they **need** to be public? If not they should be protected?

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_Timer.c
```diff
@@ -0,0 +1,69 @@
...
+		{
+			timer = new CF_ManagedTimer(managedInstance, function, interval, params);
+			return timer;
+		}
+
+		timer = new CF_Timer(instance, function, interval, params);
+		return timer;
+	}
+
+	void SetParams(CF_TimerParam params)
```
Arkensor: Is this an internal function or intended for others to use on an already created timer? If internal prefix with _

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,169 @@
...
+	private ref CF_TimerBase m_Next;
+	private CF_TimerBase m_Prev;
+	private bool m_IsActive;
+	private bool m_Destroy;
+
+	//! It is recommended to not modify these variables while the timer is active/running.
+	float m_Interval; // Seconds
+	float m_DeltaTime; // Seconds
+	float m_TimeElapsed; // Seconds
+	int m_TimeElapsedMS; // Milliseconds
```
Arkensor, line 15: Do they **need** to be public? If not they should be protected, especially if you want to avoid people setting the values ... then I would only expose them via a getter.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
```diff
@@ -0,0 +1,169 @@
...
+		m_DeltaTime += dt;
+
+		if (m_DeltaTime >= m_Interval)
+		{
+			OnTick(m_DeltaTime);
+			m_DeltaTime = 0;
+		}
+	}
+
+	static void _Update(float dt)
```
Arkensor: Just for readability, maybe rename this to `_UpdateAll` or something else that implies that this does something for all active timers. It looks a bit strange in DayZGame.c to "update a single static timer".

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+
+- The param function arguments are nulled after the first call
+- Timer running and iteration is based soley on function calls and dynamic arrays, these are expensive instructions in Enforce Script
+- The creation and management of timers can be convuluted. They were designed for UI and delayed calls. 
+
+There is one major issue with `ScriptCallQueue`, after a server has been up for a prolonged time there will be a dramatic loss of server performance, as evident from [Game Labs](). This is not just from modders forgetting to clean up their timers.
+
+The timers Community Framework is providing will offer improved performance by removing unneccessary function calls and removing the use dynamic arrays. As testing has shown with [Expression's](../ExpressionVM/index.md) there is a huge performance gain when using dual Linked Lists over even static arrays. 
+
+All timers have safeguards built in, such as with [Instance Timers](#Instance-Timers) only allowing to be ran in `Managed` classes.
```
Arkensor: Outdated, since we now also allow unsafe instances.

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+
+There are three issues with vanilla timers, they are:
+
+- The param function arguments are nulled after the first call
+- Timer running and iteration is based soley on function calls and dynamic arrays, these are expensive instructions in Enforce Script
+- The creation and management of timers can be convuluted. They were designed for UI and delayed calls. 
+
+There is one major issue with `ScriptCallQueue`, after a server has been up for a prolonged time there will be a dramatic loss of server performance, as evident from [Game Labs](). This is not just from modders forgetting to clean up their timers.
+
+The timers Community Framework is providing will offer improved performance by removing unneccessary function calls and removing the use dynamic arrays. As testing has shown with [Expression's](../ExpressionVM/index.md) there is a huge performance gain when using dual Linked Lists over even static arrays.
```
Arkensor, line 11: Linked list has no meaning to the context of "what are timers" - it's an implementation detail.
I'd remove this.

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+
+Why should you use these timers over the countless options avaliable in vanilla? Performance and reliability.
+
+There are three issues with vanilla timers, they are:
+
+- The param function arguments are nulled after the first call
+- Timer running and iteration is based soley on function calls and dynamic arrays, these are expensive instructions in Enforce Script
+- The creation and management of timers can be convuluted. They were designed for UI and delayed calls. 
+
+There is one major issue with `ScriptCallQueue`, after a server has been up for a prolonged time there will be a dramatic loss of server performance, as evident from [Game Labs](). This is not just from modders forgetting to clean up their timers.
```
Arkensor: As per discussion in dm, remove due to lack of testing.

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
+# Timers
+
+Why should you use these timers over the countless options avaliable in vanilla? Performance and reliability.
+
+There are three issues with vanilla timers, they are:
+
+- The param function arguments are nulled after the first call
+- Timer running and iteration is based soley on function calls and dynamic arrays, these are expensive instructions in Enforce Script
```
Arkensor: If this intends to make a claim that CF_Timer has better performance than the vanilla one, this needs reproducible test results linked in a benchmark section at the bottom of the readme.

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+- The creation and management of timers can be convuluted. They were designed for UI and delayed calls.
```
Arkensor, line 9: Maybe a before and after example would be nice to show a) how CF_Timer makes things easier and b) implicitly giving a quick minimal sample of how CF_Timer works that *could* be copied and already be self-explanatory enough.

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+- Timer running and iteration is based soley on function calls and dynamic arrays, these are expensive instructions in Enforce Script
+- The creation and management of timers can be convuluted. They were designed for UI and delayed calls. 
+
+There is one major issue with `ScriptCallQueue`, after a server has been up for a prolonged time there will be a dramatic loss of server performance, as evident from [Game Labs](). This is not just from modders forgetting to clean up their timers.
+
+The timers Community Framework is providing will offer improved performance by removing unneccessary function calls and removing the use dynamic arrays. As testing has shown with [Expression's](../ExpressionVM/index.md) there is a huge performance gain when using dual Linked Lists over even static arrays. 
+
+All timers have safeguards built in, such as with [Instance Timers](#Instance-Timers) only allowing to be ran in `Managed` classes. 
+
+## Instance Timers
```
Arkensor, line 15: You do not explain what an instance timer is ...

I thought instance timers are those where you need to store the return value of CF_Timer.Create(...) somewhere ... aka keeping the instance of the timer. Whereas managed timers get a `Managed` param and thus can be forgotten about?

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Basic Example
+
+```csharp
+class SomeClass : Managed
+{
+    void SomeClass()
+    {
+        float interval = 0.5; // The timer will tick every 0.5 seconds.
+        CF_Timer.Create(this, "OnTick", interval);
```
Arkensor: Oneliner is equally easy to understand I think. If not keep it as it is.
```cs
CF_Timer.Create(this, "OnTick", 0.5); // The timer will tick every 0.5 seconds.
```

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    void OnTick(CF_TimerBase timer, float deltaTime)
+    {
+        Print("Timer has ran for " + timer.m_TimerElapsed + " seconds.");
```
Arkensor, line 56: Encourage the use of the Getter ... especially if we decide to make the member protected as per comment on those. Also making sure people understand the difference between total time and the current deltaTime
```cs
Print("Timer has ran for " + timer.GetTimeElapsed() + " seconds in total.");
```

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+    }
+};
+```
+
+Memory management is handled all for you. To create a timer you run `CF_Timer.Create` with the arguments you specify. Note that an interval of `0.0` means the timer will tick every frame.
```
Arkensor: ... if they pass in a managed instance. You should mention that quickly or else someone might glimpse over that from the example and not understand that as a requirement.

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+### Advanced Example
+
+```csharp
+class Menu : UIScriptedMenu
+{
+    static const float FADE_TIME = 0.5;
+
+    void Menu()
+    {
+        CF_Timer.Create(this, "FadeIn", 0.0, new CF_TimerParam1<Widget, float>(GetLayoutRoot(), FADE_TIME));
```
Arkensor, line 71: One idea I get from this is a `Create` overload that lacks the interval param and just calls the other one with 0.0 in that place. C# likes to offer overloads that omit parameters at the front or middle as long as the order of types still uniquely identifies the overload.

```cs
CF_Timer.Create(this, "FadeIn", new CF_TimerParam1<Widget, float>(GetLayoutRoot(), FADE_TIME));
```

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    void FadeIn(CF_TimerBase timer, float deltaTime, Widget widget, float timeToFade)
```
Arkensor, line 74: Looking at this example I think it would be nicer to combine the invoke info into one small struct 

```cs
class CF_TimerInfo
{
	CF_TimerBase timer;
	float deltaTime;
}
```

So that the timer only gets
```cs
void FadeIn(CF_TimerInfo info, Widget widget, float timeToFade)
```

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+## Generic Timers
+
+Generic Timers are used in places where you want to use the same code for the timer in multiple places at once, or you want to use the timer outside of a `Managed` class or in a static/global function.
+
+### Basic Example
+
+```csharp
+class SomeClass : Managed
```
Arkensor: > you want to use the timer outside of a `Managed` class

And the example you show uses a `Managed` class? Copy-paste mistake maybe?

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+        
+        // Start this timer automatically.
+        Start();
+    }
+
+    override void OnTick(float dt)
+    {
+        Print("Timer has ran for " + m_TimerElapsed + " seconds and the owner is " + m_SomeClass + ".");
+
+        // Experiment removing the `Managed` class from `SomeClass` inheritance.
```
Arkensor: What is this comment about?

docs/Timers/index.md
```diff
@@ -0,0 +1,155 @@
...
+        {
+            Destroy();
+        }
+    }
+};
+```
+
+### Advanced Example
+
+Comparing to the advanced example from `Instance Timers`, this is how the new code will be created and ran.
```
Arkensor: I do not understand the difference to the other example. Whatever you mean with "Instance Timers" is not self-explanatory from the example below.

Arkensor (changes requested):
I do not understand the concept of what an "Instance Timer" is supposed to be. If this is about the managed vs class problem or if you mean something like a pre-built timer that can be instantiated multiple times with different params or something ... Read through the comments and whatever it is, make sure it's explained better in the docs.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_ManagedTimer.c
Jacob-Mango: Enforce Script doesn't support changing Constructor signature from inherited classes

docs/Timers/index.md
Jacob-Mango: You said in DM's to encourage the use of member variables

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_Timer.c
Jacob-Mango: Leaving the possibility open if a modder wants to update the params

docs/Timers/index.md
Arkensor: I did? Can't remember to have said that, and especially not for a context like that.
Especially with something like this, the inner workings should not be exposed. See my comment on protecting the members to avoid setting it from external code.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_DelayCall.c
Jacob-Mango: Enforce Script bug

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_DelayCall.c
Jacob-Mango: > I am not sure what the vision is for DelayCall ... the name suggests delaying ONE function call to a later time ... aka the same that ScriptCallQueue already does.

Yeah. According to the source that I've outlined in private messages to you, ScriptCallQueue has issues when a large wait time has been inputted. 

> Do you maybe mean RepatedCall/ReapeatableCall since we need to store the result of CF_DelayCall::Create() somewhere already?

FYI, that's what `CF_Timer` is doing already

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_DelayCall.c
Jacob-Mango: > Does it need managed here? Same discussion we had about normal timers - if we are responsible to handle the case when the instance that we should invoke becomes invalid.
Changing it to class would allow this to be used with existing custom classes that use inheritance.

My mistake in forgetting about `CF_DelayedCall`

> Change the delay param to be float to fit the rest of the timer functions.

Safer to use an int, since `CF_DelayedCall` is designed to operate over a very long time. `CF_Timer` uses a float because the interval between each call is expected to be within a safe range for floating-point errors it is fine. `CF_DelayedCall` is expected to only call the function once over minutes or hours later, this can have floating-point errors if a modder wanted a precise time.

JM/CF/Scripts/3_Game/CommunityFramework/Timer/CF_TimerBase.c
Jacob-Mango: We decided in private messages to encourage the use of accessing the variables over using the methods for performance reasons.

docs/Timers/index.md
Jacob-Mango: An implementation detail yes, but it acknowledges the advantages and reasonings to further strengthen the argument to use `CF_Timer` over vanilla `Timer`.

docs/Timers/index.md
Jacob-Mango: Enforce Script complains about ambiguity ;(

PR #56, opened 2021-06-16
Model View ViewModel (MVVM)

Jacob-Mango:

PR #72, opened 2021-08-16
TypeConverters

Jacob-Mango:
Merge after #58 and #74 

```csharp
class SomeClass
{
	string variable = "wow";
	
	void SomeClass()
	{
		Print(variable); // "wow"

		// creating a typeconverter that writes to a string variable
		CF_TypeConverter converter = CF_TypeConverters.Create(string);

		converter.SetBool(true);
		
		// first parameter is owning class of the variable
		// second parameter is the name of the variable it is writing to
		converter.Write(this, "variable");

		Print(variable); // "true"
	}
};
```


JM/CF/Scripts/1_Core/CommunityFramework/CF_LogLevel.c
```diff
@@ -0,0 +1,11 @@
+enum CF_LogLevel
+{
+	ERROR	= 0x000001,
+	WARNING	= 0x000010,
+	INFO 	= 0x000100,
+	DEBUG	= 0x001000,
+	TRACE	= 0x010000,
+
+	NONE	= 0,
+	ALL		= 0x011111,
```
Arkensor, line 10: Log levels are usually exclusive and not a bit mask ?!?

JM/CF/Scripts/3_Game/CommunityFramework/TypeConverter/CF_TypeConverter.c
```diff
@@ -0,0 +1,172 @@
...
+	bool FromTypename(Class instance, int index)
+	{
+		#ifdef CF_TRACE_ENABLED
+		CF_Trace trace(this, "FromVariable", "" + instance, "" + index);
+		#endif
+		CF_Log.Error("Override FromVariable!");
+		return false;
+	}
+
+	void ToVariable(Class instance, string variable)
```
Arkensor: I'd call this something like `Apply`/`ApplyTo` because it is a bit more explanatory of what's happening ...

```cs
conveter.ApplyTo(inst, "var");
Print(var); //yay
```

JM/CF/Scripts/3_Game/CommunityFramework/TypeConverter/CF_TypeConverterDate.c
```diff
@@ -0,0 +1,41 @@
+class CF_TypeConverterDate : CF_TypeConverterManaged
+{
+	private string m_Format = "YYYY-MM-DD hh:mm:ss";
```
Arkensor: Fixed format? Ehhh ... I guess no way around it?

JM/CF/Scripts/3_Game/CommunityFramework/TypeConverter/CF_TypeConverters.c
```diff
@@ -0,0 +1,90 @@
+static autoptr CF_TypeConverters g_CF_TypeConverters = null;
```
Arkensor: Is this only for debugging via trace?

JM/CF/Scripts/3_Game/CommunityFramework/Utils/CF_Localiser.c
```diff
@@ -0,0 +1,209 @@
...
+		if (m_Translates[realIndex])
+		{
+			value = Widget.TranslateString("#" + value);
+			if (value == "" || value[0] == " ")
+				return m_Strings[realIndex];
+		}
+		return value;
+	}
+
+	/* Backwards compatbility - to be removed after CF 1.4 */
```
Arkensor, line 197: Please open an issue that references this code so we remember to address that.

Arkensor (commented):
A lot of boilerplate code, only skipped over most of it, trusting that you tested your concept well.

JM/CF/Scripts/3_Game/CommunityFramework/TypeConverter/CF_TypeConverterDate.c
Jacob-Mango: For this PR, yes.

In the future, `CF_Date` may store its format but that is currently undecided.

JM/CF/Scripts/3_Game/CommunityFramework/TypeConverter/CF_TypeConverter.c
Jacob-Mango: It is named as it is to keep consistent with `FromVariable`

JM/CF/Scripts/3_Game/CommunityFramework/TypeConverter/CF_TypeConverters.c
Jacob-Mango: Yes

JM/CF/Scripts/3_Game/CommunityFramework/TypeConverter/CF_TypeConverter.c
Arkensor: You read From(a)Variable, which makes sense. But it only makes sense to Apply(To/On) or Write another variable.

Maybe Read() and Write() are better terms for this anyway. ToVariable sounds like you create a new variable instance from something like you do with ToList, ToString, ToWhatever.

Jacob-Mango: I'll rename to `Read` and `Write`

JM/CF/Scripts/3_Game/CommunityFramework/Utils/CF_Localiser.c
Jacob-Mango: https://github.com/Jacob-Mango/DayZ-Community-Framework/issues/78

docs/TypeConverters/index.md
```diff
@@ -0,0 +1,148 @@
+# Type Converters
+
+Provides a unified way of converting types of values to other types, as well as for accessing values and supporting custom serialzation.
```
Arkensor: The original sentence felt like a little stroke. Not sure what you wanted to say. I propose maybe this (?):
> Provides a unified way of converting values between different variable types and supports custom serialization using CF_IO.

docs/TypeConverters/index.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Using Type Converters
+
+Accessing a type converter retrieves the same instance always.
```
Arkensor: So you want to say
> Each type converter is a singleton instance per variable type.

docs/TypeConverters/index.md
```diff
@@ -0,0 +1,148 @@
...
+	void SetID(int id)
+	{
+		auto converter = CF_TypeConverter.Get(string);
+
+		converter.SetInt(id);
+
+		m_ID = converter.GetString();
+	}
+}
+```
```
Arkensor, line 36: Maybe explain the example with a short sentence like:

> This example shows how the type converter is used to represent an integer value as a string for internal storage

or something like that ...

docs/TypeConverters/index.md
```diff
@@ -0,0 +1,148 @@
...
+The variable name can be used to read and write the data. This is made with script reflection in mind. Just pass in the instance holding the variable and the variable name to the `Read` and `Write` methods. 
+
+```csharp
+class SomeClass
+{
+	string m_ID;
+
+	int GetID()
+	{
+		auto converter = CF_TypeConverter.Get(string);
```
Arkensor, line 47: Would it make sense to offer a Get overload that accepts the string instance m_ID and just calls Get() with the typename of the instance? that way you could switch out m_ID from string to float or what ever and the rest of the code could still work

docs/TypeConverters/index.md
```diff
@@ -0,0 +1,148 @@
...
+Unlike with normal type converters, 'Write' does not need to be called as the custom type converter written does not change the instance, just the variable within the instance.
+
+```csharp
+auto converter = CF_TypeConverter.Get(PlayerStatBase);
+converter.Read(GetGame().GetPlayer(), "m_StatDiet");
+
+Print(converter.GetFloat()); // will print out the current float variable associated with the player
+converter.SetString("0.5");
+Print(converter.GetFloat()); // will print out "0.5" as a float
+```
```
Arkensor: For somebody that knows what he wants to do and just looks up the syntax maybe this works, but I have no idea in what situation I would want to create my own type converter and I don't know how to work with the m_Value, that is just in the code with no other explanation.

If I have my own data structs like SomeState that is essentially just an int stored as m_State, how do i convert between that and int/float/string?
```
class SomeState
{
	int m_State;
}
```
If I have to manually tell which member variable to access like you do here:
> converter.Read(GetGame().GetPlayer(), "m_StatDiet");

then what does type converter even do for me?

Arkensor (commented):
Did not view any of the files I reviewed in #74

Not really sure what to say about the converter code until I understood what it is supposed to do. Type converter docs need more elaboration for dummies like me.

docs/TypeConverters/index.md
Jacob-Mango: > If I have to manually tell which member variable to access like you do here:

> converter.Read(GetGame().GetPlayer(), "m_StatDiet");

> then what does type converter even do for me?

It's the type of feature that wouldn't be used outside of #56 and #73

docs/TypeConverters/index.md
Jacob-Mango: That's what I get for basically copying the C# docs

docs/TypeConverters/index.md
Jacob-Mango: Added `CF_TypeConverter::Get(instance, variableName)` in latest commit. But it is an O(n) operation - where `n` is the number of variables inside of `instance`

docs/TypeConverters/index.md
Arkensor: For the record: @Jacob-Mango and I decided in a conversation to keep the docs surface level for now as its primary use is with a different feature that will explain the use-case better. We might want up update the docs for this later on to make the feature of type converters understandable on its own.

Arkensor (approved):

PR #74, opened 2021-08-16
Added CF_Trace, CF_Log, CF_Date, CF_Localiser

Jacob-Mango:


JM/CF/Scripts/1_Core/CommunityFramework/CF_LogLevel.c
```diff
@@ -0,0 +1,11 @@
+enum CF_LogLevel
+{
+	ERROR	= 0x000001,
+	WARNING	= 0x000010,
+	INFO 	= 0x000100,
+	DEBUG	= 0x001000,
+	TRACE	= 0x010000,
+
+	NONE	= 0,
+	ALL		= 0x011111,
```
Arkensor: Why is there an ALL option. Log levels should be exclusive and not a bitmask to be combined.

Arkensor (commented):
Not sure why log levels are implemented as bitmask ... That is rather odd

Also without docs and all the other code from other branches, it is not easy to give feedback. I will only re-review after docs were added so I can understand the core additions of this.

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
```diff
@@ -0,0 +1,165 @@
...
+			count += value.Substring(index, 1) == char;
+		}
+		return count;
+	}
+
+	/**
+	 * @brief	Returns the number of times the specified character exists in the string
+	 * 
+	 * @param char The character that will be matched
+	 * @param firstIndex The index of the first character
```
Arkensor: Rename `firstIndex` to `firstOccurrence` to make more clear what it is.

Also add `[out]` suffix to `@param` -> `@param[out] firstOccurrence`

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
```diff
@@ -0,0 +1,165 @@
...
+	/**
+	 * @brief	Fills the string with the specified character at the start so the string matches the length. 
+	 * 			Does nothing if string is equal to or exceeds the specifiying length.
+	 * 
+	 * @param length The total string length 
+	 * @param padChar The character that will be padded
+	 * 
+	 * @code
+	 * 	CF_String str = "55";
+	 * 	Print(str.PadStringEnd(4, "0"));
```
Arkensor: copy paste error, wrong function called in example

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
```diff
@@ -0,0 +1,165 @@
...
+	 * 	>> '0055'
+	 * @endcode
+	 */
+	string PadString(int length, CF_String padChar)
```
Arkensor: Maybe call this one PadStringFront? So it matches PadStringBack?

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
```diff
@@ -0,0 +1,165 @@
...
+	 * @param length The total string length 
+	 * @param padChar The character that will be padded
+	 * 
+	 * @code
+	 * 	CF_String str = "55";
+	 * 	Print(str.PadStringEnd(4, "0"));
+	 * 	>> '5500'
+	 * @endcode
+	 */
+	string PadStringEnd(int length, CF_String padChar)
```
Arkensor: PadStringBack would be a nicer name I think because it's the "back" of the string ...

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
```diff
@@ -0,0 +1,165 @@
...
+	 */
+	string SpliceString(int start, CF_String splice)
+	{
+		int middle = start + splice.Length();
+		int end = value.Length();
+		string a = value.Substring(0, start);
+		string b = splice;
+		string c = value.Substring(middle, end - middle);
+		return a + b + c;
+	}
```
Arkensor: What happens if the splice replacement is bigger than the oiginal?
Does it work with empty splice string and index 0 and N-1?

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
```diff
@@ -0,0 +1,165 @@
...
+			}
+		}
+
+		return newString;
+	}
+
+	/**
+	 * @brief	Replaces the string of length at the index with the splice. 
+	 * 
+	 * @param start The first character to be removed
```
Arkensor: I would say `The index at which to start changing the string`, borrowed from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
```diff
@@ -0,0 +1,165 @@
...
+	 * @param padChar The string that will replace those characters
```
Arkensor: I would say `The string the content will be replaced with`

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
```diff
@@ -0,0 +1,165 @@
...
+	 */
+	string SpliceString(int start, int length, CF_String splice)
+	{
+		int middle = start + length;
+		int end = value.Length();
+		string a = value.Substring(0, start);
+		string b = splice;
+		string c = value.Substring(middle, end - middle);
+		return a + b + c;
+	}
```
Arkensor: Same thing here as with the other overload. Did you validate the edge cases?

docs/Localiser/index.md
```diff
@@ -0,0 +1,47 @@
...
+A localising system for networked messages. 
+
+## Creating and Using
+
+The following example is using strings found in the DayZ stringtable. 
+
+The localiser class has overloaded the array assignment operator to allow setting of the string table keys and placing in other variables at the desired format.
+
+```csharp
+CF_Localiser localiser = new CF_Localiser("STR_DATE_FORMAT");
```
Arkensor: Put a comment on this at the end with the value of `STR_DATE_FORMAT` so you can see where the placeholders are and understand the example better.

docs/Localiser/index.md
```diff
@@ -0,0 +1,47 @@
+# Localiser
+A localising system for networked messages.
```
Arkensor: Better:
> A utility to construct multi-part localised strings which can also be sent across the network

Because the primary function of this is not the networking ... they are 90% as useful in local use only. Sending them via RPC is just a neat thing you can also do. We discussed this via DM, but I think the network bit could/should be its own mini util built on top of `Localiser`. Right now people are supposed to send their own raw ScriptRPC according to the docs, which feels meh.

docs/Localiser/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+```csharp
+void OnRPC(PlayerIdentity sender, Object target, int rpc_type, ParamsReadContext ctx)
+{
+	CF_Localiser localiser;
+	CF_Localiser.Read(rpc, localiser);
+	Print(localiser.Format());
+	// English: "Fri, Dec 2021"
+}
+```
```
Arkensor: It would be better if this was part of the before mentioned util built on top and that a user does not need to interact as low as sending their own RPC. We should put this behind nicer looking facade.

docs/Logging/index.md
```diff
@@ -0,0 +1,77 @@
+# Logging
+
+The following table lists the possible log level values and it's corresponding function. Only information with a level equal to or higher than `CF_Log.Level` will be outputted to the `script.log` file.
```
Arkensor, line 3: It would be nice if we could connect logging to a log sink which is either scripts.log or rpt, or a custom file, or whatever else someone might make, like an HTTP logger. 

We can do this as add-on later on, but maybe keep this in mind for now and see if it was possible to implement later without breaking existing usage.

docs/Logging/index.md
```diff
@@ -0,0 +1,77 @@
...
+| INFO      | 2     | CF_Log.Info     | Logs that output information that may be relevant to the user.                                                                                                                                |
+| WARNING   | 3     | CF_Log.Warn     | Logs that highlight an abnormal action, but does cause the user to notice anything different.                                                                                                 |
+| ERROR     | 4     | CF_Log.Error    | Logs that highlight when the current flow of execution is stopped due to a failure. This should indicate if the current activity has a failure and will not result in the game shutting down. |
+| CRITICAL  | 5     | CF_Log.Critical | Logs that describe an unrecoverable event and will most likely lead to the shutdown of the game.                                                                                              |
+| NONE      | 6     |                 |                                                                                                                                                                                               |
```
Arkensor: I do not understand the idea behind the current colum setup. Why not just Delete Name, keep Level and rename function to `Enum` or `Name`

Also, NONE should have a description like `Disables all logging.`

docs/Logging/index.md
```diff
@@ -0,0 +1,77 @@
...
+SCRIPT       : [TRACE]   -SomeClass<6ec92cd0>::SomeClass Time: 0.0098ms
+SCRIPT       : [TRACE]   +SomeClass<6ec92cd0>::SomeFunction (0.5)
+SCRIPT       : [TRACE]    +SomeClass<6ec92cd0>::SomeFunction ("This is a test string", NULL)
+SCRIPT       : [TRACE]     +SomeClass<6ec92cd0>::MethodThree ()
+SCRIPT       : [TRACE]     -SomeClass<6ec92cd0>::MethodThree Time: 0.0019ms
+SCRIPT       : [TRACE]    -SomeClass<6ec92cd0>::SomeFunction Time: 0.3468ms
+SCRIPT       : [TRACE]   -SomeClass<6ec92cd0>::SomeFunction Time: 0.7035ms
+SCRIPT       : [TRACE]  -::FunctionOne Time: 1.6316ms
+SCRIPT       : [TRACE] -SomeClass<6ec92cd0> Time: 1.6046ms
+```
```
Arkensor, line 123: The output should be a sub-section of the example. Just put it below the code, with a small heading Output and you can put the note under the example output as "footnote"

docs/Logging/index.md
```diff
@@ -0,0 +1,77 @@
...
+
+		MethodThree();
+	}
+
+	void MethodThree()
+	{
+		auto trace = CF_Trace_0(this, "MethodThree");
+	}
+}
+```
```
Arkensor: I think its a great feature, but we need to make people understand it more easily.
This example alone does not tell me what to do, and I don't really understand it.

We need an explanation of each of the functions available. What does X do and when to use X for things like:
- CF_Trace_Instance
- CF_Trace_N
Also explain what CF_Trace_1 vs CF_Trace_2 is.

And you should mention that `auto trace = ` is required to be present.

Maybe when all of this is explained beforehand, this example makes more sense. I think it will probably not be enough content to justify splitting it into many pages. I think one "longer" page to explain the basics first and then some step-by-step examples from very basic to more advanced use cases might be the right call here.

docs/Date/index.md
```diff
@@ -0,0 +1,36 @@
...
+A simple utility class for calculating time differences and rudimentry outputting.
+
+## Comparision
+
+Two date objects can be compared.
+
+### Example
+
+```csharp
+CF_Date dateA = CF_Date.CreateDateTime(2021, 12, 2, 10, 24, 3);
```
Arkensor, line 11: Put a comment next to it with the datetime written out, so that people who are used to the freedom format, do not get scared.

docs/Date/index.md
```diff
@@ -0,0 +1,36 @@
...
+CF_Date dateB = CF_Date.CreateDateTime(2021, 12, 4, 10, 24, 3);
+Print(dateB.Compare(dateA).GetDay());
+// "2"
```
Arkensor, line 14: I am not sure what the intent here is? Comparing dates should usually return some kind of TimeSpan, as in they differ by 2 Days or something. Returning a new date and call GetDay there feels super strange. It's comparing apples to oranges.

Maybe have a look at how C# handles DateTime and TimeSpan. But it's a supermassive feature. I am not sure if you wanted to touch that here.

Arkensor (changes requested):
CF_Date Compare needs some major attention or needs to be scrapped. If the Compare feature was not new and I just noticed in this rename, then it needs work at a later time.

CF_Trace needs better explanations so users understand when and how to use it.

Some minor naming/description suggestions.

Details in the comments.

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
Jacob-Mango: Will add the following on 138

```
		if (middle >= end)
		{
			return a + b;
		}
```

docs/Logging/index.md
Jacob-Mango: > I do not understand the idea behind the current colum setup. Why not just Delete Name, keep Level and rename function to Enum or Name

Name and Function are both useful for modders to know. So they both know the function the loglevel is associated with and the name of the corresponding loglevel.

docs/Date/index.md
Jacob-Mango: I will remove Compare from this PR and add it back in a later PR, it is a new feature but not required for any other PRs.

docs/Localiser/index.md
Jacob-Mango: > they are 90% as useful in local use only

Judging from Expansions existing usecase for `StringLocaliser`, 70% in Server to Client RPC use. 30% local use.

> We discussed this via DM, but I think the network bit could/should be its own mini util built on top of Localiser. Right now people are supposed to send their own raw ScriptRPC according to the docs, which feels meh.

We can circle back to this after the new RPC is implemented.

docs/Logging/index.md
Jacob-Mango: Create an issue for this

docs/Localiser/index.md
Jacob-Mango: I will remove the documentation for now until CF_RPC is implemented. I will keep the feature though.

docs/Logging/index.md
Arkensor: Here it is #80

docs/Logging/index.md
Jacob-Mango: Rewrote and added mini documentation before the big example.

docs/Localiser/index.md
```diff
@@ -8,40 +8,11 @@ The following example is using strings found in the DayZ stringtable.
 The localiser class has overloaded the array assignment operator to allow setting of the string table keys and placing in other variables at the desired format.
 
 ```csharp
-CF_Localiser localiser = new CF_Localiser("STR_DATE_FORMAT");
+CF_Localiser localiser = new CF_Localiser("STR_DATE_FORMAT"); // "%a, %b %d"
```
Arkensor: What is this supposed to be? a b d? Format uses %1 %2 etc

docs/Localiser/index.md
```diff
@@ -8,40 +8,11 @@ The following example is using strings found in the DayZ stringtable.
...
 localiser.Add(5);
 localiser[1] = "Hello";
-localiser[2] = "STR_NOVEMBER";
+localiser[2] = "STR_NOVEMBER"; // "Dec"
```
Arkensor: November is December?!?

JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Trace.c
```diff
@@ -21,12 +21,19 @@ class CF_Trace
...
+		
+		string instanceName;
+		g_Script.CallFunction(instance, "GetDebugNameNative", instanceName, null);
+		if (instanceName == string.Empty)
+		{
+			instanceName = "" + instance;
+		}
 
-		return "" + instance;
+		return instanceName;
```
Arkensor: Where do these changes come from now? What is GetDebugNameNative?

Jacob-Mango: > GetDebugNameNative

It's a function that is specifically native to the `Object` class. 

> Where do these changes come from now? 

For CF_Trace to work with #72 and #56 I had to move it to `1_Core`. `Object` doesn't exist in `1_Core`.

docs/Localiser/index.md
Jacob-Mango: It's not meant to work? That's what the dayz stringtable uses.

I haven't tested the example, just searched through the dayz stringtable to find a string with formatting and then made sure it compiled.

Arkensor: Find a tested working example then :)

PR #82, opened 2021-12-10
trace, log bugfixes

Jacob-Mango:


JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Log.c
```diff
@@ -55,6 +55,16 @@ class CF_Log
...
+
+		string dump = "";
+		DumpStackString(dump);
+		array<string> outputs = new array<string>();
+		dump.Split("\n", outputs);
+
+		for (int i = 1; i < outputs.Count(); i++)
+		{
+			Print("\t" + outputs[i]);
+		}
```
Arkensor, line 67: Hide behind a param if the stack shall be outputted as well. I think default should be false and only include the PrintFormat

JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Log.c
```diff
@@ -65,5 +75,15 @@ class CF_Log
...
```
Arkensor, line 87: s. above comment

Arkensor (changes requested):

JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Log.c
Jacob-Mango: Can't add behind a param, should I add two new methods called `TraceError` and `TraceCritical`?

I would prefer Error and Critical just always output to the stack log though. This is because as the docs say these two log methods are only used when execution is halted. So the user and modder should always know the cause.

Arkensor: Ok fair enough, the prototype does not allow for that to be easily added.
Leave it as it is then.

Arkensor (dismissed):

Arkensor (approved):

PR #84, opened 2022-01-03
Created file class

Jacob-Mango:
A utility for looping through files. Also features a handful of utility functions.

```csharp
CF_Log.Level = CF_LogLevel.INFO;

array<ref CF_File> files;
if (CF_Directory.GetFiles("$profile:CommunityOnlineTools/Items/*", files))
{
	foreach (auto file : files)
	{
        CF_Log.Info("Found file: %1" + file.GetDebugName());

		CF_File copy;
   		if (file.Copy(file.GetDirectory() + file.GetFileNameWithoutExtension() + "-copy" + file.GetExtension(), copy))
    	{
        	CF_Log.Info("Copied file to: %1" + copy.GetDebugName());
    	}
	}
}
```

### DIFF
--- a/JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_Directory.c
+++ b/JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_Directory.c
@@ -5,7 +5,7 @@ class CF_Directory
 	 * @param [inout] files An array where the list of files will be written to
 	 * @param flags If it should be looked in directories or .pak files, or both
 	 * 
-	 * @return If the pattern was valid
+	 * @return If any files were found that matched the pattern
 	 */
 	static bool GetFiles(string pattern, inout array<ref CF_File> files, FindFileFlags flags = 2)
 	{
--- a/JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_Directory.c
+++ b/JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_Directory.c
@@ -51,4 +51,56 @@ class CF_Directory
 		
 		return true;
 	}
+
+	/**
+	 * @brief Creates all directories and subdirectories in the specified path unless they already exist. 
+	 * 
+	 * @param directory The directory to create.
+	 * 
+	 * @return If the directories could be created
+	 */
+	static bool CreateDirectory(string directory)
+	{
+		array<string> directories();
+		string path = directory;
+
+		string previousPath = "";
+		while (true)
+		{
+			previousPath = path;
+			if (FileExist(previousPath))
+			{
+				break;
+			}
+			
+			path = CF_Path.GetDirectoryName(path);
+			if (previousPath == path)
+			{
+				path = previousPath;
+				break;
+			}
+			
+			if (previousPath[previousPath.Length() - 1] == CF_Path.FILESYSTEM_IDENTIFIER)
+			{
+				return false;
+			}
+			
+			directory = CF_Path.GetFileNameEx(previousPath, path);
+			directory = directory.Substring(0, directory.Length() - 1);
+
+			directories.Insert(directory);
+		}
+
+		for (int i = directories.Count() - 1; i >= 0; i--)
+		{
+			path = path + directories[i] + CF_Path.DIRECTORY_SEPARATOR;
+			
+			if (!MakeDirectory(path))
+			{
+				return false;
+			}
+		}
+
+		return true;
+	}
 };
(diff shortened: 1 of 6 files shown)


JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_File.c
```diff
@@ -0,0 +1,167 @@
...
+	protected bool m_IsValid;
+
+	/**
+	 * @param pattern The search pattern
+	 * @param [inout] files An array where the list of files will be written to
+	 * @param flags If it should be looked in directories or .pak files, or both
+	 * 
+	 * @return If the pattern was valid
+	 */
+	static bool Find(string pattern, inout array<ref CF_File> files, FindFileFlags flags = 2)
```
Arkensor: Can't we do `FindFileFlags flags = FindFileFlags.ALL`?

JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_File.c
```diff
@@ -0,0 +1,167 @@
...
+			files = new array<ref CF_File>();
+		}
+
+		string fileName;
+		FileAttr fileAttr;
+		FindFileHandle handle = FindFile(pattern, fileName, fileAttr, flags);
+		if (!handle)
+		{
+			return false;
+		}
```
Arkensor: Please double check that handle is only NULL when the pattern was invalid. Try to search for a file that does not exist and see if handle is valid. I suspect it might be NULL as well, in which case we can not tell if the pattern was correct or not. In that case have the function return void.

JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_File.c
```diff
@@ -0,0 +1,167 @@
...
+			file = new CF_File();
+			file.m_FileName = fileName;
+			file.m_IsDirectory = fileAttr & FileAttr.DIRECTORY;
+			file.m_IsHidden = fileAttr & FileAttr.HIDDEN;
+			file.m_IsReadOnly = fileAttr & FileAttr.READONLY;
+			file.m_IsValid = !(fileAttr & FileAttr.INVALID);
+			files.Insert(file);
+
+			more = FindNextFile(handle, fileName, fileAttr);
+		}
```
Arkensor: Turn it into a while(true)

`if(!FindNextFile(handle, fileName, fileAttr)) break;` on L50

Arkensor (commented):
Missing docs or use case in PR description?

JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_File.c
Jacob-Mango: Compile error unfortunately

JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_File.c
Jacob-Mango: Returns false with invalid pattern.

Using the below code with pattern "$profile:CommunityOnlineTools/*" and "$profile:SomeFolderDoesntExist/*", the first returns true, the second returns false.

```
array<ref CF_File> files;
string pattern = "$profile:CommunityOnlineToolcs/*";

bool success = CF_File.Find(pattern, files);
Print(success);
foreach (auto file : files)
{
	Print(file.GetDebugName());
}
```

Jacob-Mango:
All functions were tested (and succeed) to work in the intended ways after `CreateDirectory` was added.

Arkensor (commented):
No docs for CF_File or CF_Directory? Or do you think the inline docs there are sufficient?

JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_Directory.c
```diff
@@ -0,0 +1,106 @@
+class CF_Directory
+{
+	/**
+	 * @param pattern The search pattern
+	 * @param [inout] files An array where the list of files will be written to
+	 * @param flags If it should be looked in directories or .pak files, or both
+	 * 
+	 * @return If the pattern was valid
```
Arkensor: Tested and no the return value has nothing to say about if the pattern itself was valid. FindFile returns a NULL handle if the pattern was valid but there was no file that matched it. So instead this line should be changed to:

`* @return If any files were found that matched the pattern`

Arkensor (approved):

PR #86, opened 2022-01-05
Feature/surface

Jacob-Mango:
```csharp
auto surface = CF_Surface.At(GetGame().GetPlayer().GetPosition());
surface = CF_Surface.At(GetGame().GetPlayer().GetPosition()); // if the player didn't move, then the surface will be of the same instance
Print(surface.Interior); // check if we are in an interior surface
```


Jacob-Mango:
Tested in Expansion for getting vehicle surface information. Works.

Arkensor (commented):
I did not know this kind of stuff was needed but looks like it could save someone a bit of pain so alrighty ...

I would suggest you change the constructors of the Surfaces to protected so they can only be constructed once (singleton) internally. Also I would then recommend to make the class members constant to avoid someone modifying the instances?

Jacob-Mango:
> I would suggest you change the constructors of the Surfaces to protected so they can only be constructed once (singleton) internally.

I'll have to make it `private` as Enforce Script produces a compile error with it protected. Still allows people to mod the class which will be much more likely than a modder creating a new class.

> Also I would then recommend to make the class members constant to avoid someone modifying the instances?

Unfortunately, that will result in a compile error.

Arkensor:
> > Also I would then recommend to make the class members constant to avoid someone modifying the instances?
> 
> Unfortunately, that will result in a compile error.

I think you can do do `const string m_Value = GetValue();` or I think also `const string m_Value = valueVariable;` that both execute when the class is being created.

The idea would be to store the values in static buffers which can be private to the class. Then have a static CF_Surface.Create(...) method where you accept all the values as parameters. Assign them to the buffers, then create a real `new CF_Surface` instance where its normal `public const` members are assigned with the buffer values. It appears a bit complicated at first, but I think this way would work and allow us to have const variables that we assign once that are not compile time constants.

If the error you got was unrelated to assigning cost variables let me know which one it was.

Jacob-Mango:
You never "completed" the previous review so consider this the notification to re-review

Arkensor (dismissed):

PELange (approved):

Arkensor (approved):

PR #87, opened 2022-01-06
Rewrote module system, added documentation

Jacob-Mango:
The module system was re-written to still interact neatly with the old module system and prevent conflicts while enjoying some needed optimization. Now not every module is looped when an event is fired as they have to manually be enabled beforehand.

Below is an example module which will send a message to the player 5 seconds after they log into the server.

```csharp
[CF_RegisterModule(SomeModule)]
class SomeModule : CF_ModuleWorld
{
	override void OnInit()
	{
		super.OnInit();

		// Enable the 'OnInvokeConnect' event
		EnableInvokeConnect();
	}

	override void OnInvokeConnect(Class sender, CF_EventArgs args)
	{
		super.OnInvokeConnect(sender, args);

		auto player = CF_EventPlayerArgs.Cast(args).Player;

		// Send the messsage 20 seconds later, this event can fire before the HUD is loaded and so the message won't be seen
		GetGame().GetCallQueue(CALL_CATEGORY_SYSTEM).CallLater(this.SendMessage, 20000, false, player, "Welcome to the server!");
	}

	void SendMessage(PlayerBase player, string message)
	{
		// Send a message by firing the RPC to the player
		GetGame().RPCSingleParam(player, ERPCs.RPC_USER_ACTION_MESSAGE, new Param1<string>(message), true, player.GetIdentity());
	}
};
```


Jacob-Mango:
The new module system works as intended.

Stil reverse compatible with mods (Community Online Tools) using the old module system.

Jacob-Mango:
I have decided to use `CF_EventArgs` instead of raw parameters. The reasoning is so the function signature won't complain about compile errors if they ever have to be changed. BI *could* rework the login system again for all we know so this is forward planning just in case.

Jacob-Mango:
Tested again after new changes, module system works as intended.

And it is still reverse compatible with mods using the old system such as Community Online Tools.

docs/Modules/index.md
```diff
@@ -0,0 +1,83 @@
+# Modules
+
+## Base Classes
+
+Depending on which `ScriptModule` you desire the module to be in, you change the inheritance to allow for more events. All the possible events in the modules are documented within the the respective classes.
+
+| Script Module | CF Module        |
+| ------------- | ---------------- |
+| 1_Core        | `CF_ModuleCore`  |
+| 2_GameScript  | `CF_ModuleCore`  |
```
Arkensor, line 10: Twice core?

docs/Modules/index.md
```diff
@@ -0,0 +1,83 @@
...
+| 3_Game        | `CF_ModuleGame`  |
+| 4_World       | `CF_ModuleWorld` |
+| 5_Mission     | `CF_ModuleWorld` |
```
Arkensor, line 13: Twice World? No Mission module? I mean even if it did nothing you could still maybe make one and have it be a 1:1 inherit from World or does that not make much sense?

docs/Modules/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Events
+
+All events within the modules are by default disabled. Once enabled the event can't be disabled. To enable an event you call the corresponding method starting with 'Enable'. 
+
+| Event Method                              | Enable Method                 | When Fired?                                        | Executing Platform |
+| ----------------------------------------- | ----------------------------- | -------------------------------------------------- | ------------------ |
+| `CF_ModuleCore::OnMissionStart`           | `EnableMissionStart`          | Called when `MissionBase::OnMissionStart` is ran   | ALL                |
+| `CF_ModuleCore::OnMissionFinish`          | `EnableMissionFinish`         | Called when `MissionBase::OnMissionFinish` is ran  | ALL                |
+| `CF_ModuleCore::OnMissionLoaded`          | `EnableMissionLoaded`         | Called when `MissionBase::OnMissionLoaded` is ran  | ALL                |
```
Arkensor, line 23: Is it really "is ran"? I thought it was "is run" or "was ran" but not a mix of it possible. Not sure if that is British vs American English or something.

Arkensor (approved):

docs/Modules/index.md
Jacob-Mango: The same base module class can be used for either scriptmodule

docs/Modules/index.md
Jacob-Mango: The same base module class can be used for either scriptmodule.

Didn't want to add a new class when there was no real reason to due to the '#InitGlobals' issue.

PR #92, opened 2022-01-21
Fixed CTD when closing the game

Jacob-Mango:
Also;

* fixed some minor typos
* added trace logs everywhere
* remove some bad refs where it won't compile error


JM/CF/Scripts/1_Core/CommunityFramework/Module/CF_ModuleCoreManager.c
```diff
@@ -51,9 +51,14 @@ class CF_ModuleCoreManager
 	static void _OnDestroy()
 	{
 #ifdef CF_TRACE_ENABLED
-		auto trace = CF_Trace_0("Create");
+		auto trace = CF_Trace_0("CF_ModuleCoreManager", "_OnDestroy");
 #endif
```
Jacob-Mango, line 56: This is the CTD fix

Arkensor (approved):
This stuff better be tested to not have broken something ...
Please no more prs that change multiple things. One for fixing the crash. one for removing refs. one for adding logging to all the classes. If we have to revert one of the changes in here, we can't .. we will need to undo everything.

PR #94, opened 2022-01-23
Bugfix/inconsitencies

Jacob-Mango:
Added small bits of inline documentation that doesn't need a PR

Removed the disabled logging and stat tracking from RPCManager

Moved OnMissionLoaded from game and server to MissionBase

Removed disabled permission references

Fixed RPCManager compile error with VPP

### DIFF
--- a/JM/CF/Scripts/5_Mission/CommunityFramework/Mission/MissionBase.c
+++ b/JM/CF/Scripts/5_Mission/CommunityFramework/Mission/MissionBase.c
@@ -1,6 +1,6 @@
 modded class MissionBase
 {
-	protected bool m_CF_MissionLoaded = false;
+	protected bool m_bLoaded = false;
 	
 	void MissionBase()
 	{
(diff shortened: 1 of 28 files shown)


JM/CF/Scripts/5_Mission/CommunityFramework/Mission/MissionBase.c
```diff
@@ -1,20 +1,35 @@
 modded class MissionBase
 {
+	protected bool m_bLoaded = false;
+
```
Arkensor, line 4: This should be `m_cf_bLoaded` as someone else might add a variable with this generic name into mission base.

Arkensor (commented):
Small naming conflict avoidance if possible. It existed before, but we could get rid of it when we are already working on this class

JM/CF/Scripts/5_Mission/CommunityFramework/Mission/MissionBase.c
Jacob-Mango: Can't fix it right now. Some mods will have a compile error

Arkensor (approved):

PR #102, opened 2022-02-05
Added a method for programmatically toggling trace function

Jacob-Mango:
```csharp
static bool TRACE_METHODONE = false;

void MethodOne(float pDt)
{
		auto trace = CF_Trace_1(TRACE_METHODONE, this, "MethodOne").Add(pDt);
}
```

### DIFF
--- a/JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Trace.c
+++ b/JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Trace.c
@@ -28,11 +28,11 @@ class CF_Trace : Managed
 		}
 		
 		string instanceName;
-		g_Script.CallFunction(instance, "GetDebugNameNative", instanceName, null);
-		if (instanceName == string.Empty)
-		{
+		//g_Script.CallFunction(instance, "GetDebugNameNative", instanceName, null);
+		//if (instanceName == string.Empty)
+		//{
 			instanceName = "" + instance;
-		}
+		//}
 
 		return instanceName;
 	}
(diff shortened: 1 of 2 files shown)


docs/Logging/index.md
```diff
@@ -62,9 +62,23 @@ class ClassTwo
...
+```csharp
+class ClassTwo
+{
+	static void MethodFour(string variableOne, int variableTwo)
+	{
+		auto trace = CF_Trace_2("ClassTwo", "MethodFour").Add(variableOne).Add(variableTwo);
+	}
+}
+```
+
```
Arkensor: I think you wanted to not copy paste the example above and add the variable you were talking about here

Arkensor (changes requested):

JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Trace.c
```diff
@@ -27,11 +28,11 @@ class CF_Trace : Managed
...
 		string instanceName;
-		g_Script.CallFunction(instance, "GetDebugNameNative", instanceName, null);
-		if (instanceName == string.Empty)
-		{
+		//g_Script.CallFunction(instance, "GetDebugNameNative", instanceName, null);
+		//if (instanceName == string.Empty)
+		//{
 			instanceName = "" + instance;
-		}
+		//}
```
Arkensor, line 35: Why not just wrap this in a `if(g_Script)` and or `if(instance)` to avoid nullptr problems when called and either of them is not available yet? Now we lack the GetDebugNameNative information even if it is available.

Jacob-Mango: It's not crashing because g_Script is null. It's crashing due to some change behind the scenes in 1.16 for `GetDebugNameNative`

Arkensor (approved):

PR #103, opened 2022-02-05
Added persistence info to avoid read on first boot of CF after installation

Jacob-Mango:
This does not support unloading CF from the mods. Once loaded it can't be removed.

The data is written to `$mission:storage_{instanceId}\communityframework\cf_modstoragemodule.bin`

It is a raw binary file where it just stores the persistent IDs of the entities continuously to the one file. Once added it is not removed. As a result, this system would only break if DayZ were to ever get a shared hive between servers. 

Even if an entity were to be deleted and a new entity generated with the same ID it wouldn't matter as the ID will always be registered when the entity has its storage saved. 

The point of failure is if a persistent ID was meant to be saved but somehow didn't get saved. At that point, there would be additional data from modstorage that would fail to not get loaded. 

This system will only work when the server updates to 1.16. CF must get updated before 1.16 reaches stables otherwise we increment the version and this system is although added, delayed until the next version patch.


JM/CF/Scripts/3_Game/CommunityFramework/ModStorage/CF_ModStorageModule.c
```diff
@@ -0,0 +1,165 @@
...
+			MakeDirectory(folder);
+		}
+
+		folder += "communityframework/";
+		if (!FileExist(folder))
+		{
+			MakeDirectory(folder);
+		}
+
+		m_FilePath = folder + "cf_modstoragemodule.bin";
```
Arkensor: I find the name is a bit odd. sure we prefix our classes etc, but i think this file could have a nicer name such as `modstoragemeta.bin` or `modstorageids.bin` or simply `modstorage.bin`

JM/CF/Scripts/3_Game/CommunityFramework/ModStorage/CF_ModStorageModule.c
```diff
@@ -0,0 +1,165 @@
...
+		{
+			FileHandle handle = OpenFile(m_FilePath, FileMode.READ);
+			if (handle != 0)
+			{
+				// Reads 4 ints (16 bytes) at a time
+				int data[4];
+				while (ReadFile(handle, data, 16) > 0)
+				{
+					_AddEntity(data[0], data[1], data[2], data[3], true);
+				}
```
Arkensor, line 112: Clever!

JM/CF/Scripts/3_Game/CommunityFramework/ModStorage/CF_ModStorageModule.c
```diff
@@ -0,0 +1,165 @@
...
+		if (!map_b3)
+		{
+			map_b3 = new map<int, ref Param1<bool>>();
+			map_b2[b3] = map_b3;
+		}
+
+		auto val_b4 = map_b3[b4];
+		if (!val_b4)
+		{
+			val_b4 = new Param1<bool>(loaded);
```
Arkensor, line 158: Do we actually need a param1<bool>? I mean for serialization it comes down to just being a bool. Why don't we just write bool directly? We do not have anything smaller to store than a bool/int do we?

JM/CF/Scripts/4_World/CommunityFramework/ModStorage/CF_ModStorageObject.c
```diff
@@ -16,31 +18,44 @@ class CF_ModStorageObject<Class T> : CF_ModStorageBase
...
 
 	override void OnStoreSave(ParamsWriteContext ctx)
 	{
-#ifdef CF_TRACE_ENABLED
+#ifdef CF_MODSTORAGE_TRACE
 		auto trace = CF_Trace_1(this, "OnStoreSave").Add(ctx);
 #endif
 
-		if (GetGame().SaveVersion() < 116)
+		if (GetGame().SaveVersion() < CF_ModStorage.GAME_VERSION_FIRST_INSTALL)
```
Arkensor, line 31: This was our preparation update we did a while ago? So and data earlier than that we do not attempt to write anything right?
Or what version is `116`? I don't remember the numbering scheme for it.

JM/CF/Scripts/4_World/CommunityFramework/ModStorage/CF_ModStorageObject.c
```diff
@@ -52,27 +67,43 @@ class CF_ModStorageObject<Class T> : CF_ModStorageBase
...
+			if (!m_Module.IsEntity(b1, b2, b3, b4))
+			{
+				// Since the entity wasn't found we can assume that CF is freshly loaded
+				// Highly unlikely anything else happened that can cause this
+				// * A new entity won't take the ID of an old entity if it is in circulation
+				// * If this is a new entity, OnStoreSave would've been called at some point before 'OnStoreLoad'
+				return true;
+			}
+		}
+
```
Arkensor, line 98: So essentially we are saying, if the entity is not in the bin file, we do not attempt to read the `cf_version` from its buffer. After the first save of that entity with CF installed it should be in the file, hence has at least the `cf_version` data written to it, so it is safe to do the next step. And once we know the `cf_version` we can know what else we are allowed to read?

And all we need the dayz update for is to say "alright existing installations do not have the file yet, so we need to read the `cf_version` because it should already be in the data" and after that happend one time they have their files and everything is fine. So we use the `version` incrementing as "trigger once" for our upgrade process?

The name `GAME_VERSION_WIPE_FILE` confused me a little. Not sure what this is supposed to mean. Version `127` will be the next major stable update?

Arkensor (commented):
Have you tested if creating / reading the file has increase the loading times significantly? How long would it take for lets say 250.000 persistent id reads that we might have added to a big server? I know you posted a benchmark some time, but I don't have the numbers anymore.

Jacob-Mango:
> Have you tested if creating / reading the file has increase the loading times significantly? 

Saving will have no discenerable effect on the performance. Reading the file will be slow but it only happens once on the first entity load or save. 

> How long would it take for lets say 250.000 persistent id reads that we might have added to a big server?

Not too long. Roughly an additional 6 CPU ticks per entity best case. There would be no major difference with worst-case in regards to time and instead, it would just be the memory layout of the 4 rows of IDs.

JM/CF/Scripts/4_World/CommunityFramework/ModStorage/CF_ModStorageObject.c
Jacob-Mango: IIRC storage version 116 corresponds to game version 1.10

JM/CF/Scripts/4_World/CommunityFramework/ModStorage/CF_ModStorageObject.c
Jacob-Mango: > So essentially we are saying, if the entity is not in the bin file, we do not attempt to read the cf_version from its buffer. After the first save of that entity with CF installed it should be in the file, hence has at least the cf_version data written to it, so it is safe to do the next step. And once we know the cf_version we can know what else we are allowed to read?

Correct

> And all we need the dayz update for is to say "alright existing installations do not have the file yet, so we need to read the cf_version because it should already be in the data" and after that happend one time they have their files and everything is fine. So we use the version incrementing as "trigger once" for our upgrade process?

Correct

> The name GAME_VERSION_WIPE_FILE confused me a little. Not sure what this is supposed to mean. Version 127 will be the next major stable update?

Correct, storage version 127 corresponds to game version 1.16

JM/CF/Scripts/3_Game/CommunityFramework/ModStorage/CF_ModStorageModule.c
Jacob-Mango: Can change. Was holdover from before the file was generated inside of the `communityframework` folder.

JM/CF/Scripts/3_Game/CommunityFramework/ModStorage/CF_ModStorageModule.c
Jacob-Mango: Just saving a variable for the sake of it. It could be anything.

Using a Param1<> so I can check for null without running a `Contains` call because it just made the code look a little bit cleaner.

Arkensor (approved):

PR #106, opened 2022-02-15
Bugfix/modstorage disable

Jacob-Mango:


JM/CF/Scripts/4_World/CommunityFramework/ModStorage/CF_ModStorageObject.c
```diff
@@ -33,11 +34,13 @@ class CF_ModStorageObject<Class T> : CF_ModStorageBase
 			return;
 		}
 
+#ifndef CF_MODSTORAGE_MODULE_DISABLE
 		int b1, b2, b3, b4;
 		m_Entity.GetPersistentID(b1, b2, b3, b4);
 
 		// Add the entity to the file so on next load the game knows that it can read the modstorage for the entity
 		m_Module.AddEntity(b1, b2, b3, b4);
+#endif
```
Arkensor, line 43: This define block is inside what is disabled already with `CF_MODSTORAGE_DISABLE`?

Arkensor (changes requested):

JM/CF/Scripts/4_World/CommunityFramework/ModStorage/CF_ModStorageObject.c
Jacob-Mango: So if you decide to enable ModStorage in the future, you can do so with the Persistence file being disabled

Arkensor (approved):

PR #111, opened 2022-02-19
Fixed memory leak

Jacob-Mango:
Imagine saying you will no longer be working on DayZ mods and then making a pull request just a little over a week later on one. Couldn't be me.

### DIFF
--- a/JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Trace.c
+++ b/JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Trace.c
@@ -202,310 +202,372 @@ class CF_Trace : Managed
 
 static CF_Trace CF_Trace_Instance(Class instance)
 {
-	return new CF_Trace(CF_Trace.FormatInstance(instance), string.Empty, 0, true);
+	auto trace = new CF_Trace(CF_Trace.FormatInstance(instance), string.Empty, 0, true);
+	return trace;
 }
 
 static CF_Trace CF_Trace_0(Class instance, string stackName)
 {
-	return new CF_Trace(CF_Trace.FormatInstance(instance), stackName, 0, true);
+	auto trace = new CF_Trace(CF_Trace.FormatInstance(instance), stackName, 0, true);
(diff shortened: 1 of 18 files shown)


JM/CF/Scripts/3_Game/CommunityFramework/Game/DayZGame.c
```diff
@@ -1,5 +1,7 @@
 modded class DayZGame
 {
+	ref CF_EventRPCArgs m_CF_EventRPCArgs = new CF_EventRPCArgs();
+
```
Arkensor, line 4: Shouln't this also be autoptr?

JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Trace.c
```diff
@@ -202,310 +202,372 @@ class CF_Trace : Managed
 
 static CF_Trace CF_Trace_Instance(Class instance)
 {
-	return new CF_Trace(CF_Trace.FormatInstance(instance), string.Empty, 0, true);
+	auto trace = new CF_Trace(CF_Trace.FormatInstance(instance), string.Empty, 0, true);
+	return trace;
```
Arkensor, line 206: What is the idea behind this?

JM/CF/Scripts/3_Game/CommunityFramework/Game/DayZGame.c
Jacob-Mango: It works as-is with no reports of memory leaks. The problem may have been passing new instances straight into a function without the instance ever being set to a ref variable.. Never gets a chance for it's ref count to be updated when the function call ends.

JM/CF/Scripts/1_Core/CommunityFramework/Logging/CF_Trace.c
Jacob-Mango: Assigning to a variable so it can have it's ref count set.

Arkensor (approved):

PR #115, opened 2022-04-04
Use a different approach to check entity for modstorage

lava76:
Only needs to keep track of players. See comments in CF_ModStorageModule.c near the top.


lava76:
Important note: The wipefile version has been bumped to 129 (1.17 Experimental). If you are on 1.17 Exp and have existing storage, just add an empty `modstorageplayers.bin` to your `storage_<x>/communityframework` folder _before_ starting the server with modstorage for the 1st time and it'll work fine.

lava76:
~Turns out identity is NULL at a point where I'm checking it, this went unnoticed because it doesn't seem to cause any readily visible problems. I'll try and come up with a fix for this later.~ (fixed by 9c4e571)

lava76:
Just as a heads-up, this version of ModStorage is used by Expansion since the 14th by duplicating some of the bits to make it work until CF catches up

Arkensor (dismissed):
If a version of this has been working flawlessly in your other mods, and this change respects backwards compatibility of loading old approach data (which i think it does right?) then its good to be merged.

lava76:
Yes it does. Current plan is to target DayZ 1.27 for ModStorage being turned active in CF.

Arkensor (approved):

PR #124, opened 2022-12-02
Add read/write byte array to IO

wrdg:


wrdg:
Updated ReadBytes to reserve array size ahead of insertion

Jacob-Mango (approved):

PR #130, opened 2023-02-11
Limit CF_OnUpdate rate for perf boost

lava76:
CF_OnUpdate loop is fairly expensive especially on server. Limit update rate to 25 ms on server. Significant performance boost was observed when many modules are using update loops (e.g. COT gained 10% tick rate increase on server with this change).


JM/CF/Scripts/5_Mission/CommunityFramework/Mission/MissionBase.c
```diff
@@ -1,5 +1,9 @@
 modded class MissionBase
 {
+#ifdef SERVER
+	float m_CF_UpdateTime;
+#endif
```
Arkensor, line 5: Does `SERVER` define work for singleplayer e.g. COM? What if you disconnect afterward and connect to a server? I would suspect the code is preprocessed once on the game compilation, later role changes not affecting it.

lava76: `SERVER` is only defined when running a dedicated server.

Arkensor: So this solution does not work in single-player? If so that is not an option. We can't rely on that define then.

JM/CF/Scripts/5_Mission/CommunityFramework/Mission/MissionBase.c
```diff
@@ -23,6 +23,17 @@ modded class MissionBase
 
 	void CF_OnUpdate(float timeslice)
 	{
+#ifdef SERVER
```
Arkensor, line 26: How is timeslice "inaccurate"? It's supposed to be used that way, it is used in all of the game's code. Besides floating arithmetic errors that can happen from time to time, I don't see any reason why to make additional function calls.

Also, this code does not anticipate lag spikes, it needs to accumulate updates and then
```
while(accumulator >= 0.025)
{
    DoSomething();
    accumulator -= 0.025;
}
```

So that any update logic behind it can rely on a max update rate of 0.025 for simulation constraints.

Arkensor (changes requested):

JM/CF/Scripts/5_Mission/CommunityFramework/Mission/MissionBase.c
lava76: Rate limiting is not supposed to be used on client or in SP, it only makes sense on a dedicated server. On client, the update loop may also be used for rendering, thus should go as fast as it can.

JM/CF/Scripts/5_Mission/CommunityFramework/Mission/MissionBase.c
lava76: `timeslice` not being accurate in the DayZGame update loop specifically (I'm not aware it's inaccurate in EOnSimulation for example) is a DayZ bug. CF_OnUpdate already accumulates the time with the proposed, that's what m_CF_UpdateTime and the delta are used for.

Arkensor: Yes, you accumulate but only flush once with delta times that can be bigger than 0.025. I thought the intention was a fixed simulation timestamp. If you wanted to limit the max refresh rate only I guess that works.

> timeslice not being accurate in the DayZGame update loop specifically is a DayZ bug.

In what way? What bug? If there would be a bug the entire game would be affected by it. If so it would be patched in the base game anyway, so no need to waste resources in CF code to bypass an intended information source. Give steps to reproduce on the fb tracker. For CF code please revert back to simply using the delta time that is given by the game.

lava76: > I thought the intention was a fixed simulation timestamp. If you wanted to limit the max refresh rate only I guess that works.

Yes. It will be fixed in the sense that when a CF OnUpdate loop finishes within the allotted time, it will not go above 25 ms. There needs to be allowance for that though, as time taken for one loop can never be guaranteed.

> In what way?

Timeslice is roughly 1.5 to 2 times lower than real time due to rounding and float precision. This means if you use timeslice to limit rate, it'll run 1.5 to 2 times faster than intended.

> If there would be a bug the entire game would be affected by it. If so it would be patched in the base game anyway, so no need to waste resources in CF code to bypass an intended information source.

I don't think this particular issue is fixable in the base game due to rounding and floating point precision limits. It doesn't really matter for the CF side anyway. The GetGame().GetTickTime() call doesn't add any measurable overhead (GameLabs/CF Tools do the same to facilitate measuring OnUpdate tick rate, which would not be accurate if they were to rely on the timeslice).

Arkensor:
There was a reason why the defines were in their own file / addon some time. Not sure if this is still needed. Please clarify why it was moved. @Jacob-Mango do you remember why we had it there?

lava76:
Relevant commit on devel 317f91fceea012c54e6e309d5201a928860d2323

PR #140, opened 2023-02-15
Use ToAscii instead of Hash

wrdg:
Stops using Hash to get a single character string ascii value, and uses ToAscii as this is now linked properly in script vm.

### DIFF
--- a/JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_BinaryWriter.c
+++ b/JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_BinaryWriter.c
@@ -25,7 +25,7 @@ class CF_BinaryWriter : CF_IO
 
 	override void WriteChar(string value)
 	{
-		m_Stream.AppendCurrent(value.Hash());
+		m_Stream.AppendCurrent(value.ToAscii());
 	}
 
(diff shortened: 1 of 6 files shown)


JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_BinaryWriter.c
```diff
@@ -94,15 +94,15 @@ class CF_BinaryWriter : CF_IO
 		
 		for (int i = 0; i < value.Length(); i++)
 		{
-			m_Stream.AppendCurrent(value.Substring(i, 1).Hash());
+			m_Stream.AppendCurrent(value[i].ToAscii());
```
Arkensor: Why would @Jacob-Mango have used substring for this before instead of direct array access? I would think this updated code is correct but I'd like to confirm with him.

Arkensor: > Get(n) calls substr(1), and [n] is just (simplified) find and replace for Get(n)
So substr(1) would be faster

is what he said to me via Discord. So please undo the Substring replacements @wrdg

Arkensor (approved):

PR #146, opened 2023-04-29
Add ReadAllLines, ReadAllText, and ReadAllBytes to CF_File

InclementDab:


JM/CF/Scripts/1_Core/CommunityFramework/Files/CF_File.c
```diff
@@ -320,6 +320,110 @@ class CF_File : Managed
...
+		while (byte) 
+		{
+			fileContents += byte.AsciiToString();
+			byte = fileStream.Next();
+		}
+		
+		fileStream.Close();
+		
+		return fileContents;
+	}
```
Arkensor, line 351: I think for performance reasons something like this should not be read through the whole file io pipeline with every single character going through multiple script calls? Wouln't `FGets` be an order of magnitude faster?

What was your use case when you came up with the idea to add these 3 methods?

PR #150, opened 2023-07-19
Refactor/date

TrueDolphin:
I have implemented a new static function, CF_Date::CompareDates, within the CF_Date class. This function allows for easy comparison of two date objects in the CF_Date format. 

Example use:
```c++    
int a = 60; //a = offset in minutes
CF_Date date1 = CF_Date.Now(true);
CF_Date date2 = CF_Date.Epoch(Birthday + a);
int hoursDiff, minutesDiff;
CF_Date.CompareDates(date2, date1, hoursDiff, minutesDiff);

if (hoursDiff > -1) {
    if (minutesDiff > -1) {
        return true;
    }
}
```

### DIFF
--- a/JM/CF/Scripts/3_Game/CommunityFramework/Utils/CF_Date.c
+++ b/JM/CF/Scripts/3_Game/CommunityFramework/Utils/CF_Date.c
@@ -283,10 +283,10 @@ class CF_Date : Managed
 		return iTimestamp;
 	}
 
-    static int Compare(CF_Date date1, CF_Date date2)
+	int Compare(CF_Date other, bool useUTC = false)
 	{
-		int timestamp1 = date1.GetTimestamp();
-		int timestamp2 = date2.GetTimestamp();
+		int timestamp1 = CF_Date.Now(useUTC).GetTimestamp();
+		int timestamp2 = other.GetTimestamp();
 
 		if (timestamp1 < timestamp2) {
 			return -1;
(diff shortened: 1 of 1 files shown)


JM/CF/Scripts/3_Game/CommunityFramework/Utils/CF_Date.c
```diff
@@ -283,6 +283,15 @@ class CF_Date : Managed
 		return iTimestamp;
 	}
 
+    static void CompareDates(CF_Date date1, CF_Date date2, out int hours, out int minutes)
```
Arkensor: A better name might be just Compare and then return -1 0 +1 for the default java comparator idea so you know if the dates are equal, the first is smaller or the second one is smaller.

Arkensor (changes requested):

JM/CF/Scripts/3_Game/CommunityFramework/Utils/CF_Date.c
TrueDolphin: so you'd prefer?:
```c#
    static int Compare(CF_Date date1, CF_Date date2)
	{
		int timestamp1 = date1.GetTimestamp();
		int timestamp2 = date2.GetTimestamp();

		if (timestamp1 < timestamp2) {
			return -1;
		} else if (timestamp1 > timestamp2) {
			return 1;
		} else {
			return 0;
		}
	} 
````

Arkensor: Yeah, that looks better. Though it can be made nonstatic and only accept one param to compare against. 
Maybe for your original purpose of getting the time difference make an additional `void CalculateDifference(CF_Date other, out int hours, out in minutes)`

TrueDolphin: adjusted and added utc bools for compatibility with CF_Date::now

Arkensor (approved):

PR #158, opened 2024-02-10
Update batch files

Jacob-Mango:
Updated workflow batch files to move away from 3rd party tool dependency


JM/CF/Workbench/Logs/Build.deploy
Arkensor, line 1: Logs should probably not be part of the git?

Arkensor (changes requested):

PR #159, opened 2024-03-09
Add instance validation check

babiyvadim:


Arkensor:
Normally the instance should not be nulled and it should especially not be an invalid pointer. The wrapper and the map structure around it is strong referenced. It would probably help if you shared the circumstances of the original crash you seemed to have?

babiyvadim:
Hi Arkensor! I've just checked my methods to reproduce crashes and noticed that one of them doesn't reproduce. 
```
class TestInstance
{
    void TestInstance()
    {
        GetRPCManager().AddRPC("RPC_TestInstance", "TestRPC", this, SingeplayerExecutionType.Client);
    }

    void TestRPC(CallType type, ref ParamsReadContext ctx, ref PlayerIdentity sender, ref Object target)
    {
        Print("RPC received");
    }
}


modded class MissionGameplay
{
    ref TestInstance test_instance;
    override void OnMissionStart()
    {
        super.OnMissionStart();
        // Create a new instance of the TestInstance class
        test_instance = new TestInstance();
        // Destroy the instance when the function ends
        delete test_instance;
    }
}

modded class MissionServer
{
    override void InvokeOnConnect(PlayerBase player, PlayerIdentity identity)
    {
        super.InvokeOnConnect(player, identity);
        GetGame().GetCallQueue(CALL_CATEGORY_SYSTEM).CallLater(SendRPC, 2000, false, player);
    }

    void SendRPC(PlayerBase player)
    {
        Print("Sending RPC");
        GetRPCManager().SendRPC("RPC_TestInstance", "TestRPC", null, true, player.GetIdentity());
    }
}
```

This one still can crash the game, but that's because TestInstance class isn't inherited from Managed class. When I made this merge request Managed instances also didn't work properly. My suggestion is to add an option to register/deregister RPC dynamically for more flexibility or usage. It can be useful in cases when you want to receive some data from a server only in instances of classes like UIScriptedMenu (which can be created and destroyed at any time)

Arkensor:
I see, then I think the better approach here would be to offer a `GetRPCManager().RemoveRPC("RPC_TestInstance", "TestRPC", this)` method so you can properly deregister them. But the code works as it should. It will of course be deleted if you forcefully dispose the instance. The ref will not do anything in this case. It would worked most likely if you did not manually call delete but let the reference counting dispose your managed instance :)

If you add a removerpc method and test it I can merge that.

PR #162, opened 2024-04-05
Minor fixes and ability to remove registered RPC from RPC Manager

babiyvadim:
Hi, @Arkensor!
I've added a new method as you requested and noticed some minor issues with a CF_String class. 
image
That's how CF_Date.Format returned value before the fix

### DIFF
--- a/JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
+++ b/JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
@@ -108,7 +108,7 @@ class CF_String : string
 	string PadStringFront(int length, CF_String padChar)
 	{
 		string newString = value;
-		length = newString.Length() - length;
+		length = length - newString.Length();
 
 		if (length > 0)
 		{
@@ -137,7 +137,7 @@ class CF_String : string
 	string PadStringBack(int length, CF_String padChar)
 	{
 		string newString = value;
-		length = newString.Length() - length;
+		length = length - newString.Length();
 
 		if (length > 0)
 		{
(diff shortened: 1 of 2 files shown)


JM/CF/Scripts/3_Game/CommunityFramework/RPC/RPCManager.c
```diff
@@ -185,6 +185,24 @@ class RPCManager
...
+		auto trace = CF_Trace_2(this).Add(modName).Add(funcName);
+		#endif
+
+		if ( m_RPCActions.Contains( modName ) )
+		{
+			if ( m_RPCActions[ modName ].Contains( funcName ) )
+			{
+				m_RPCActions[ modName ].Remove( funcName );
+				return true;
+			}
```
Arkensor, line 201: If you remove something you do not need to check if it existed. Remove should probably also return a bool if was successful or not.

Afterwards a check for `m_RPCActions[modName].Count() == 0` to also do `m_RPCActions.Remove(modName)` might make sense to clean it up.

Arkensor (changes requested):

JM/CF/Scripts/1_Core/CommunityFramework/CF_String.c
```diff
@@ -108,7 +108,7 @@ class CF_String : string
 	string PadStringFront(int length, CF_String padChar)
 	{
 		string newString = value;
-		length = newString.Length() - length;
+		length = length - newString.Length();
```
Arkensor, line 111: Please open seperate reviews for seperate issues. Remove the change from this PR's branch.
That change needs an example of how it behaves right now and how it does after the change.

JM/CF/Scripts/3_Game/CommunityFramework/RPC/RPCManager.c
babiyvadim: Default map.Return method doesn't return an execution status, that's why it is wrapped in if operator. I'll change it from bool to void

PR #169, opened 2024-05-22
Fixed an issue with XML reader not ignoring "-" character for tags

Arkensor:


Jacob-Mango (approved):
Approving my own commit

PR #170, opened 2024-07-12
Fixed ConfigReader::IsLetterOrDigit

lava76:
Side note: After this change, `_cf_characters` in 3_Game/CommunityFramework/CommunityFramework.c is no longer used and can probably be removed


Arkensor:
If it is not used anywhere then I would say remove it. its a framework private var (_underscore) so no backwards compatibility needed

Arkensor (approved):

PR #173, opened 2025-01-24
bugfix: CF_Base64Stream decode error after 3rd character

incinatuscodes:
```
        CF_Base64Stream stream = new CF_Base64Stream();
        stream.Decode("dGhpcyBpcyBhbiBlbmNvZGVkIGJhc2U2NCwgSGVsbG8sIFdvcmxkIQ");
        CF_TextReader reader = new CF_TextReader(stream);
        Print(reader.ReadLine());
```

Does not work on the current version of CF (verified only for DayZDiag_x64.exe, might be that the production version of DayZ is okay). The output is unreadable non-ASCII after the first three characters.

See the commit message for the justification of the fix.


lava76:
Hi, thanks for your contribution. This will be in the next CF-Test update, and hit CF stable before DayZ 1.27 stable release.

incinatuscodes:
Hi @lava76, thanks. I forgot to enable Github email privacy, would you mind overwriting the commit with my users.noreply.github email? See commits under #176 for that

PR #176, opened 2025-01-24
bugfix: CF_TextReader: fix off by on error in Read functions

incinatuscodes:
the final `ReadByte()` advances the file position and triggers the `EOF()` check at the next iteration, without adding the final byte to `str`


JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_TextReader.c
```diff
@@ -114,11 +121,13 @@ class CF_TextReader : CF_IO
 	{
 		string str;
 
-		int byte = m_Stream.Next();
-		while (!CF_Encoding.IsWhitespace(byte) && CF_Encoding.IsAlphanumeric(byte) && !EOF())
+		while (!EOF())
 		{
+			int byte = m_Stream.Next();			
+			if (!CF_Encoding.IsWhitespace(byte) || !CF_Encoding.IsAlphanumeric(byte)) {
```
lava76: The IsWhitespace condition is flipped

incinatuscodes: Fixed, thanks for noticing that.

JM/CF/Scripts/1_Core/CommunityFramework/IO/CF_TextReader.c
```diff
@@ -55,12 +57,14 @@ class CF_TextReader : CF_IO
 	{
 		string str;
 
-		int byte = ReadByte();
 		bool decimal = false;
-		while (!CF_Encoding.IsWhitespace(byte) && (CF_Encoding.IsNumeric(byte) || (!decimal && byte == 46)) && !EOF())
+		while (!EOF())
 		{
+			int byte = ReadByte();
+			if ( CF_Encoding.IsWhitespace(byte) || (!CF_Encoding.IsNumeric(byte) || (decimal && byte != 46)) ) {
```
lava76: Should be `if ( CF_Encoding.IsWhitespace(byte) || (!CF_Encoding.IsNumeric(byte) && (decimal || byte != 46)) )`, but there's also a bug that existed prior to your changes, which is that decimal is never set to true after the first '.' is encountered. Full solution probably like this 
```c
			if (byte == 46)
			{
				if (decimal)
					break;
				
				decimal = true;
			}
			else if (CF_Encoding.IsWhitespace(byte) || !CF_Encoding.IsNumeric(byte))
			{
				break;
			}
```

incinatuscodes: Thanks, I incorporated that change onto this branch for you, feel free to cherry-pick it if you haven't solved it elsewhere.

PR #186, opened 2025-03-15
fix: Fix Sign() per the doc

LeMyst:
The documentation was not coherent with the code.


lava76:
The existing code was correct. Sign may never return a ~~non~~zero value, that will break things.

(edit: fixed typo)

Arkensor:
I guess the sign value of 0 is up for debate, but how it is mostly used returning 1 is the correct option. So if anything we should just fix the docs

LeMyst:
> The existing code was correct. Sign may never return a nonzero value, that will break things.

you have a double negative here 🤔

lava76:
Fixed the docs in 6812122a

PR #201, opened 2025-11-17
fix two compile errors when building the project from scratch

incinatuscodes:
In https://github.com/Arkensor/DayZ-CommunityFramework/issues/197 I began building CF locally and ran into two errors with the `production` branch of CF. I exclusively use the diag exe for testing so I am not sure whether these issues went unnoticed due to the retail version being less strict, but otherwise this PR fixes the following two errors:
CF-Bug-01
and
CF-Bug-02


lava76:
production is out-of-date, always use dev-lava as the recent branch.

incinatuscodes:
Sorry my mistake, I just used the default branch.

I'll cherry pick to your branch and edit the PR in a second

incinatuscodes:
Closing this as the first commit was duplicated, the second is still valid under #202

